PR #1, opened 2018-11-07
Fix some readme typos

danthompson:


README.md
```diff
@@ -55,8 +55,8 @@ StandardRB, don't be shocked if things change a bit!]**
 
 ## Usage
 
-ONce you've installed `standard`, you should be able to use the `standard`
-program. The simplest use case would be checking the style of all JavaScript
+Once you've installed `standard`, you should be able to use the `standard`
```
larquin: 👍

kadams54:
I noticed a few more typos; rather than open my own PR I thought maybe they could be added to this one:

"The simplest use case would be checking the style of all JavaScript files in the current working directory:"

To:

"The simplest use case would be checking the style of all Ruby files in the current working directory:"

And:

"Note: by default standard will look for all *.rb files (and some other files typically associated with Ruby like \`*.gemspec\` and \`Gemfile)"

To:

"Note: by default standard will look for all *.rb files (and some other files typically associated with Ruby like \`*.gemspec\` and \`Gemfile\`)"

Issue #16, opened 2018-11-14
overwrite --version & --help


searls:

searls:
Landed in 0.0.14

PR #19, opened 2018-11-15
Allow Standard to integrate with AtomLinter

aergonaut:
👋👋

I was really excited to see Standard come out! After reading the Readme, I thought it would be easy to use Standard with any RuboCop-compatible tools because, like the Readme says, Standard is just a wrapper around RuboCop. It didn't work at first, but I only had to make a few small changes to the gem to get it working. Hopefully these are changes that would be okay to include in the gem itself!

- The CLI now supports `-v` and `--version` to print out the RuboCop version that Standard is using under the hood. This is because AtomLinter requests to know the RuboCop version in order to alter the CLI flags it passes when linting.

  I'm not sure how this would relate to #16, or if that issue is talking about something else.
- The config file now supports `silence_cta`. Setting this to `true` will silence the CTA message that is printed after the warning output. This is because AtomLinter wants to consume the RuboCop offenses as JSON, but having the CTA message printed at the end makes the contents of stdout invalid JSON. The default of this option is `false` so it should not impact the default behavior of the tool.

  I also thought about making this a CLI flag, but I wasn't sure which one would be a better idea.

I believe these changes should enable Standard to work with other linter tools that are compatible with RuboCop besides AtomLinter. However, I am not a user of any other tools, so I can't vouch for compatibility one way or the other.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,6 +109,20 @@ $ rake standard STANDARDOPTS="--format progress"
 $ rake standard STANDARDOPTS="lib \"app/**/*\""
 ```
 
+## Using with AtomLinter
+
+Standard is compatible with [AtomLinter](https://atomlinter.github.io/)'s
+[linter-rubocop](https://github.com/AtomLinter/linter-rubocop) package. To use
+Standard with linter-rubocop, you need to do two things:
+
+1. Configure linter-rubocop to run `standard` as its **Command**.
+2. Add the flags `--silence-cta --no-display-cop-names` to the end of the
+   **Command**. The first flag will suppress Standard's pre-1.0 CTA message when
+   warnings are detected. The second will prevent RuboCop from showing cop names
+   in offense messages.
+
+Now you can see Standard's warnings in Atom.
+
 ## What you might do if you're clever
 
 If you want or need to configure Standard, there are a _handful_ of options
--- a/lib/standard/cli.rb
+++ b/lib/standard/cli.rb
@@ -11,7 +11,13 @@ module Standard
     end
 
     def run
-      rubocop_config = @config.call(@argv)
+      rubocop_config, standard_config = @config.call(@argv)
+
+      if standard_config.options[:version]
+        puts Standard::VERSION
+        return SUCCESS_STATUS_CODE
+      end
+
       runner = RuboCop::Runner.new(
         rubocop_config.options,
         rubocop_config.config_store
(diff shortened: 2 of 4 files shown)


searls:
Awesome! I think I would like to propose two changes to the behavior of this (I have not yet looked at the code):

1. I feel like this should be a CLI flag rather than a .standard.yml option, since it only applies to one context (being run by an lint editor plugin). See how we "intercept" `--fix` as an example. Is that reasonable? If so, what should that flag be named? Is it specific to atom or will any editor plugin need the same format?

2. My goal in #16 is actually to print the version of _standard_ and not RuboCop, which it currently does because we are forwarding options. So when #16 is implemented, my goal would be for `-v`/`--version` would print `0.0.10` and so forth. If the above concern introduces a flag like `--atom-editor` or `--editor`, could an invocation be combined like `--atom-editor -v` for the behavior you want?

Excuse my ignorance, as I'm not a plugin user

aergonaut:
Sounds good.

I think for the `--version` flag, the problem for Atom specifically is that linter-rubocop specifically passes `--version` to the command. It's used to determine whether or not linter-rubocop should pass `--no-display-cop-names`. It's not the end of the world if `standard --version` reports something different though. The note in the Readme could just suggest to pass `--no-display-cop-names` manually in the plugin configuration.

100% agree that it should be a CLI option instead of a config file option. I was 50/50 split on which one was best, so I'll go with your suggestion!

searls:
I'm sorry I mucked this up but I wanted to finish writing tests and refactoring the config class before accepting PRs. I'm happy to take a second look at this and reimplement it for you but it might be a week. 

If you don't mind reworking it. I'd ask to make sure it has some kind of test coverage before merging.

aergonaut:
No problem, I'll most likely be able to take a look at it this week. I think I mostly understand how the config class works now, so I think I should be able to get by.

wingrunr21:
The change to remove the CTA for the JSON formatter will be great for `vscode-ruby` as well. I'm currently detecting the last curly brace in order to strip that from the output.

searls:
@wingrunr21 I think a subsequent contribution roundaboutly obviated this PR and I failed to close it.

Because the CTA is [now part of the standard formatter](https://github.com/testdouble/standard/blob/master/lib/standard/formatter.rb#L6-L9), using the JSON formatter will have the side effect of not printing the CTA.

From the lib's own repo, running: `$ standardrb test/fixture/cli/unfixable-bad.rb -f json` will print:

```json
{"metadata":{"rubocop_version":"0.64.0","ruby_engine":"ruby","ruby_version":"2.6.0","ruby_patchlevel":"0","ruby_platform":"x86_64-darwin18"},"files":[{"path":"test/fixture/cli/unfixable-bad.rb","offenses":[{"severity":"warning","message":"Lint/AssignmentInCondition: Wrap assignment in parentheses if intentional","cop_name":"Lint/AssignmentInCondition","corrected":false,"location":{"start_line":3,"start_column":12,"last_line":3,"last_column":12,"length":1,"line":3,"column":12}}]}],"summary":{"offense_count":1,"target_file_count":1,"inspected_file_count":1}}
```

Which seems like valid JSON to me (or at least, piping it to [jq](https://stedolan.github.io/jq/) seems to work fine).

I'm going to close this unless I'm missing something here.

wingrunr21:
Ah, missed this change. Thanks!

Issue #32, opened 2018-11-29
advice: vim ale


will:
Not sure if an issue is the best place to ask, but does anyone have advice on how to hook this up with https://github.com/w0rp/ale ? It already has support for rubocop, so I imagine it's doable, but I'm bad at vimscript :(

rosston:
I haven't tested it, but `*ale-integrations-local-executables*` in [ALE's vim doc](https://github.com/w0rp/ale/blob/0a384a49d371838903d8401c5358ec60f3f4266d/doc/ale.txt) points a way forward for a short-term fix. I think you could do this in your vimrc:
```vim
let g:ale_ruby_rubocop_executable = 'standardrb'
```

will:
Thanks @rosston, that worked!

PR #33, opened 2018-11-30
Add readme section for running in editors, starting with vim

will:
ref #32

@searls let me know if this is what you were thinking, I can word it however.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,3 +310,11 @@ format: progress
 Refer to RuboCop's [documentation on
 formatters](https://rubocop.readthedocs.io/en/latest/formatters/) for more
 information.
+
+## How do I run standard in my editor?
+
+It can be very handy to know about failures while editing to shorten the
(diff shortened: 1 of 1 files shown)


searls:
LGTM - thanks!

searls:
Fixes #32

searls:
Hey Will, do you find that this actually works? It's not enough for me. Setting this and `:echo g:ale_ruby_rubocop_executable` returns `standardrb`, but each file is still getting errors from what seem to be rubocop's default settings

will:
Yeah, I do think it's working for me:

With the line in my vimrc, I get
image

Commenting it out I get
image

searls:
Huh, indeed this works fine on my second computer and not my first. Lost 45 minutes trying to answer why.

Welp.

(╯°□°）╯︵ ┻━┻

PR #36, opened 2018-12-03
Fixes README.md styles

NickLaMuro:
I mean... you had to see this coming... right?

Maybe not, I mean, I doubt the linter works on the `.md` files by default anyway, right...

```
/me checks the source of `standard` just to be sure... hold plz...
```

...

(oh... it just uses `rubocop` under the hood... well that is good)

...

```
/me checks the source of `rubocop`... hold plz...
```

...

Okay, yeah, you don't... and probably checking code blocks for this stuff would be a PITA anyway... #scopeCreep

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ No decisions to make. It just works. Here's a [⚡ lightning talk ⚡](https://w
 Install by adding it to your Gemfile:
 
 ```ruby
-gem "standard", :group => [:development, :test]
+gem "standard", group: [:development, :test]
 ```
 
 And running `bundle install`.


searls:
LOLLLLL - standard originally enforced hashrockets until I learned about the Ruby 3 kwargageddon. Good catch

NickLaMuro:
@searls Huh, interesting.  I was unaware of the Ruby 3 changes before you mentioned that.  I assume this is the ticket regarding the "kwargageddon" (_dum dum **DUM!!!**_) you are referring to?

https://bugs.ruby-lang.org/issues/14183

For those passing by and were unaware (like myself).

searls:
Yep!

PR #38, opened 2018-12-05
Disable Style/RaiseArgs

knu:
The recommended style `raise RuntimeError, "message"` generally looks better than `raise RuntimeError.new("message")` because it is a variant of `raise "message"` with an explicit exception class supplied.  However, when the argument is not a message string, the comma style does not read well because the second argument is not what `raise` would take but is just an argument for the first argument's constructor that's passed through.

e.g.
```ruby
raise ActiveRecord::RecordInvalid, record
```

So, I'm not in favor of forcing this style unless this cop learns to do the conversion only if the second argument is a string literal.

What do you think?

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -967,9 +967,6 @@ Style/PreferredHashMethods:
 Style/Proc:
   Enabled: true
 
-Style/RaiseArgs:
-  Enabled: true
-
 Style/RandomWithOffset:
   Enabled: true
 


searls:
Agree. This seems like a fight nobody was asking for.

searls:
Ideally I want the config to be as brief as possible (while still being disable-by-default and complete, so that changes to the Rubocop default ruleset do not inadvertently change Standard). If someone has a question about a specific rule, I hope they will search GitHub or `git log -S "Style/RaiseArgs"` to learn what happened previously

PR #40, opened 2018-12-07
Disable Style/PerlBackrefs

knu:
It is painful to see every occurrence of `$n` expanded to `Regexp.last_match(n)` only to get more lengthy and slower code.

For instance, the following piece of code:
```ruby
str.gsub!(/([a-z]+)_([a-z]+)/) { $1.capitalize + $2.capitalize }
```
gets corrected to this:
```ruby
str.gsub!(/([a-z]+)_([a-z]+)/) { Regexp.last_match(1).capitalize + Regexp.last_match(2).capitalize }
```
and it no longer fits nicely in one line on the screen.

I, as one coming from the Unix background, don't feel like `$1`-`$9` are so [cryptic](https://github.com/rubocop-hq/ruby-style-guide#no-perl-regexp-last-matchers) you should avoid them, but is that true for the folks out there?

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -958,9 +958,6 @@ Style/PercentLiteralDelimiters:
 Style/PercentQLiterals:
   Enabled: true
 
-Style/PerlBackrefs:
-  Enabled: true
-
 Style/PreferredHashMethods:
   Enabled: true
 


rosston:
I am not _primarily_ a Rubyist, so what's cryptic for me is probably a little different than others (although I also come from a Unix background).

I would say that, in your example with a block passed to `gsub`, the global variables _don't_ look cryptic. The block is right near the regex, so it's pretty simple to infer what the `$n` vars mean (or know what to search the internet for if you want to confirm).

However, in [the example](https://github.com/rubocop-hq/ruby-style-guide#no-perl-regexp-last-matchers) you linked to in that Ruby style guide:

```ruby
/(regexp)/ =~ string
...

# bad
process $1

# good
process Regexp.last_match(1)
```

I think the global variables are indeed rather cryptic and point to the danger in them being global variables, since they could be used in code very far from the intended regular expression.

knu:
@rosston I see your point, but the said example looks rather unfamiliar to me because you should only use `$1` if the match actually succeeds.  You normally use `$1` in a block or in a then/when clause as follows:

```ruby
case string
when /(regexp)/
  # ...
  process $1
end
# or
if /(regexp)/ =~ string
  # ...
  process $1
end
```
And I think using `$1` in a context introduced with a regexp match makes it less dangerous looking.

Also, I'd like to point out that `Regexp.last_match(1)` is a global function and no much different from `$1` in that they both look like access to global data.

knu:
Perhaps the cop should instead suggest using named back references wherever possible.
 
```ruby
if /(?<matched>regexp)/ =~ string
  process = matched
end 
```

r-obert:
I agree. I see nothing wrong with `$0-$9` and by now I would consider them to be well known although way less google-able than `Regexp.last_match`.

> It is painful to see every occurrence of $n expanded to Regexp.last_match(n) only to get more lengthy and slower code.

Right, but seems that rubocop prioritises appearance over everything else, including performance and
functionality.

For example, it has two default cops that mean `BasicObect` is basically unsupported by this style guide:

```
1. Style/NilComparison
2. Preferring is_a? (... don't know the cop name)
```

r-obert:
> I think the global variables are indeed rather cryptic and point to the danger in them being global variables, since they could be used in code very far from the intended regular expression.

Just to be clear, even though `$0-9` appear as global variables, they are thread local and cannot be re-assigned:

```ruby
$1 = 5
SyntaxError: (eval):2: Can't set variable $1
```

knu:
Yes, `$1`-`$9` are method local and thread local, and readonly, but you can still change it by reassigning `$~`.

Here's an interesting blog post recently written in Japanese: https://nacl-ltd.github.io/2018/11/08/gsub-wrapper.html ([Google Translate](https://translate.google.com/translate?hl=en&sl=auto&u=https%3A%2F%2Fnacl-ltd.github.io%2F2018%2F11%2F08%2Fgsub-wrapper.html))

searls:
How about we just disable this? It doesn't sound like either style is consistently preferable and it's not a very important point

r-obert:
Cheers it's always good to learn something new :)

PR #55, opened 2018-12-27
Output "call for feedback" message from the standard formatter

### LINKED ISSUE #53: The help message breaks parsing after using --format json
If I call standardrb with: `standardrb --format json > output.json`, the message about disagreeing with defaults is in the output at the end, which prevents a direct parse of that json file, without further preprocessing.

I'd be happy to fix this myself, but not sure the approach you'd like.

:one: STDERR for the message
:two: Don't output the message on an explicit format flag
:three: Something else?

twe4ked:
Fixes #53.

[`RuboCop::Formatter::BaseFormatter ` class](https://github.com/rubocop-hq/rubocop/blob/668f18c25d5736fe6ad7df0e8165d50c5c21dffb/lib/rubocop/formatter/base_formatter.rb) for reference.


searls:
Thanks! We need to figure out #58 before I'm comfortable merging anything, though :/

searls:
looks like there's a new conflict here. maybe resolving it will tickle circle re #58

searls:
Landed in 0.0.23 cc/ @cschneid

cschneid:
Awesome! 👏

PR #57, opened 2018-12-31
Define header_printed_already ivar to remove warnings

edwardloveall:
When running tests I got a failure because stderr printed a message:

```
warning: instance variable @header_printed_already not initialized
```

Turns out, lib/standard/formatter.rb uses the header_printed_already
instance variable but doesn't define it until late in execution. I
recreated the initialize method from RuboCop::Formatter::BaseFormatter
and defined header_printed_already to false to remove this
warning/error and fixed the test.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    standard (0.0.19)
+    standard (0.0.20)
       rubocop (>= 0.60)
 
 GEM
@@ -53,4 +53,4 @@ DEPENDENCIES
   standard!
 
 BUNDLED WITH
-   1.17.1
+   1.17.2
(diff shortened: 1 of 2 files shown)


searls:
Woah, I just realized Circle isn't running against PRs, apparently? We should probably fix that first

(My previous attempts of clearing this warning all ran into subtle rubocop failures)

edwardloveall:
Hah! It looks like you added essentially this in https://github.com/testdouble/standard/commit/3ec94476c42850d6d2a63375590ac2fa10750d34

Not sure why I didn't have that in my master but pulling/rebasing made my commit obsolete I think ¯\\\_(ツ)_/¯

As long as it works 😄

searls:
lol

PR #59, opened 2018-12-31
Add --no-fix option

### LINKED ISSUE #54: Add a way to see if changes are needed
We were looking to integrate this into our CI suite as a "is formatting needed?" check, but we couldn't figure out a clean way to do that. I think it would be useful if we could enforce the style guide by making it part of the CI flow.

Crystal has a `format` tool that uses a `--check` option for this which returns a non-zero exit code if any formatting/linting is necessary, but doesn't actually do formatting itself:

https://github.com/crystal-lang/crystal/blob/master/src/compiler/crystal/command/format.cr#L26

That's one way to go about it, but whatever the solution I think it would be useful to have a way to fail CI if formatting is needed.

edwardloveall:
In the case that you use the config to automatically fix your violations `fix: true`, this flag will override that setting in the negative direction, just as --fix overrides the lack of that setting.

I went with `no-` as the prefix because it seems to be used most commonly for inverting options.

fixes #54

### DIFF
--- a/lib/standard/merges_settings.rb
+++ b/lib/standard/merges_settings.rb
@@ -20,7 +20,7 @@ module Standard
 
     def separate_argv(argv)
       argv.partition { |flag|
-        ["--fix", "--version", "-v", "--help", "-h"].include?(flag)
+        ["--fix", "--no-fix", "--version", "-v", "--help", "-h"].include?(flag)
       }
     end
 
@@ -29,6 +29,9 @@ module Standard
         if arg == "--fix"
           cli_flags[:auto_correct] = true
           cli_flags[:safe_auto_correct] = true
+        elsif arg == "--no-fix"
+          cli_flags[:auto_correct] = false
+          cli_flags[:safe_auto_correct] = false
         end
       }
     end
--- a/lib/standard/runners/help.rb
+++ b/lib/standard/runners/help.rb
@@ -9,7 +9,7 @@ module Standard
 
           Options:
 
-            --fix             Automatically fix failures where possible
+            --[no-]fix        Automatically fix failures where possible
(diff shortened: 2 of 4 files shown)


lib/standard/runners/help.rb
```diff
@@ -9,7 +9,7 @@ def call(config)
 
           Options:
 
-            --fix             Automatically fix failures where possible
+            --[no-]fix        Automatically fix failures where possible
```
searls: Please add `--no-fix` as a separate entry, since 99.9999% of the time people should use `--fix` and this is likely to confuse people

lib/standard/merges_settings.rb
```diff
@@ -29,6 +29,9 @@ def parse_standard_argv(argv)
         if arg == "--fix"
           cli_flags[:auto_correct] = true
           cli_flags[:safe_auto_correct] = true
+        elsif arg == "--no-fix"
+          cli_flags[:auto_correct] = false
+          cli_flags[:safe_auto_correct] = false
```
searls: What happens if both are specified? I think we should have a test that last-in-wins

edwardloveall: Done in d16fd43f72c0a1d7bad83a036354c03c7d90d7a6

lib/standard/runners/help.rb
edwardloveall: Done in 62e22d268434bbd8b26fc1547108c1e337be94de

searls:
landed in 0.0.21

searls:
Oh would you mind patching the README too? Sorry!

edwardloveall:
Yeah no worries. I throw that up in another PR

PR #61, opened 2018-12-31
Add --no-fix to README

edwardloveall:
Let me know if this isn't what you had in mind 🙂 

https://github.com/testdouble/standard/pull/59#issuecomment-450678822


edwardloveall:
I used the exact config setting in the header so someone could maybe find it when searching the page.

searls:
Deal

PR #62, opened 2019-01-02
Fix a nil error in Formatter when no files were examined

cschneid:
The `@all_uncorrected_offenses` variable was only initialized in the
`file_finished` method. But if no files were analyzed, it never got
initialized, and in `finished`, it was still nil, and nil doesn't
respond to `.empty?`.

This change moves the initialization into the Formatter's initialize
method, so it's always present.


searls:
good catch, thanks!

PR #69, opened 2019-01-31
Only show run command with fix message if command was run without it

### LINKED ISSUE #60: Only print "--fix to automatically fix some problems." when some problems are fixable
We currently print  `standard: Run `standardrb --fix` to automatically fix some problems.` unnecessarily. Stop printing it when:

- [x] `--fix` was used in the run (fixed by #69)
- [ ] none of the detected problems were (safely) autofixable

Code is in the lib/standard/formatter.rb

chadmmills:
Fixes #60 

I believe this handles the first task on the issue. I'm not sure if there is something extra I need to check for handling the second task - `Stop printing it when ... none of the detected problems were (safely) autofixable`

### DIFF
--- a/lib/standard/formatter.rb
+++ b/lib/standard/formatter.rb
@@ -42,14 +42,24 @@ module Standard
 
       output.print <<-HEADER.gsub(/^ {8}/, "")
         standard: Use Ruby Standard Style (https://github.com/testdouble/standard)
-        standard: Run `#{command}` to automatically fix some problems.
       HEADER
+
+      unless auto_correct_option_provided?
+        output.print <<-HEADER.gsub(/^ {10}/, "")
+          standard: Run `#{command}` to automatically fix some problems.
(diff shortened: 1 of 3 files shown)


lib/standard/formatter.rb
```diff
@@ -42,14 +42,24 @@ def print_header_once
...
       @header_printed_already = true
     end
 
     def print_call_for_feedback
       output.print "\n"
       output.print CALL_TO_ACTION_MESSAGE
     end
+
+    def auto_correct_option_provided?
+      options[:auto_correct] || options[:safe_auto_correct]
```
chadmmills: Wasn't sure if I needed to check for both options here?

searls: Seems reasonable

searls:
Thanks for this. Gonna reopen the issue and leave the other task open

PR #77, opened 2019-02-08
Only print suggestion to run --fix if there are actually fixable cop violations

### LINKED ISSUE #60: Only print "--fix to automatically fix some problems." when some problems are fixable

emilianoLeite:
Fixes #60

## Decisions
From what I understood, there were two possible ways to tackle this problem:
1. Ask *something* whether each raised offense is autocorrectable or not;
2. Automatically run `--fix` in a clone of the verified file and check that no offense was raised.

I chose option **1.** because it seemed the simpler one.

## Result
After some (non-trivial) time digging into the Rubocop codebase, I found out that the `Offense` class has a [`attr_reader :cop_name`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/offense.rb#L52), which returns the `Cop` subclass that raised said `Offense`. This `Cop` class also detains the information of [whether the related offense is autocorrectable or not](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/autocorrect_logic.rb#L15).

After finding that out, all that was left to do was to somehow leverage that information inside `Standard::Formatter`.

I'm not sure if my current solution diverges from some community or project Best Practice ™, but I would really appreciate some feedback before I invest more time into its refactoring.

Additional comments/reasoning can be found in the commit messages and diffs comments.

### DIFF
--- a/lib/standard/formatter.rb
+++ b/lib/standard/formatter.rb
@@ -1,7 +1,10 @@
 require "rubocop"
+require_relative "helpers/autocorrectable"
 
 module Standard
   class Formatter < RuboCop::Formatter::BaseFormatter
+    include Standard::Helpers::Autocorrectable
+
     CALL_TO_ACTION_MESSAGE = <<-CALL_TO_ACTION.gsub(/^ {6}/, "")
       Notice: Disagree with these rules? While StandardRB is pre-1.0.0, feel free to submit suggestions to:
         https://github.com/testdouble/standard/issues/new
@@ -14,12 +17,12 @@ module Standard
     end
 
     def file_finished(file, offenses)
-      uncorrected_offenses = offenses.reject(&:corrected?)
-      @all_uncorrected_offenses += uncorrected_offenses
-      print_header_once unless uncorrected_offenses.empty?
+      @uncorrected_offenses = offenses.reject(&:corrected?)
+      @all_uncorrected_offenses += @uncorrected_offenses
+      print_header_once unless @uncorrected_offenses.empty?
       working_directory = Pathname.new(Dir.pwd)
 
-      uncorrected_offenses.each do |o|
+      @uncorrected_offenses.each do |o|
         absolute_path = Pathname.new(file)
         relative_path = absolute_path.relative_path_from(working_directory)
         output.printf("  %s:%d:%d: %s\n", relative_path, o.line, o.real_column, o.message.tr("\n", " "))
@@ -44,7 +47,7 @@ module Standard
         standard: Use Ruby Standard Style (https://github.com/testdouble/standard)
       HEADER
 
-      unless auto_correct_option_provided?
+      if should_suggest_auto_correct?
         output.print <<-HEADER.gsub(/^ {10}/, "")
           standard: Run `#{command}` to automatically fix some problems.
         HEADER
@@ -60,6 +63,10 @@ module Standard
 
     def auto_correct_option_provided?
       options[:auto_correct] || options[:safe_auto_correct]
+    end
+
+    def should_suggest_auto_correct?
+      !auto_correct_option_provided? && @uncorrected_offenses.any?(&method(:autocorrectable_offense?))
     end
   end
 end
--- a/lib/standard/helpers/autocorrectable.rb
+++ b/lib/standard/helpers/autocorrectable.rb
@@ -2,7 +2,8 @@ module Standard
   module Helpers
     module Autocorrectable
       def autocorrectable_offense?(offense)
-        cop_instance(offense.cop_name).support_autocorrect?
+        cop = cop_instance(offense.cop_name)
+        cop.support_autocorrect? && safely_autocorrectable?(cop)
       end
 
       def cop_instance(cop_name)
@@ -12,6 +13,11 @@ module Standard
       def constantized_cop(cop_name)
         namespaced_cop_class = cop_name.to_s.gsub("/", "::")
         RuboCop::Cop.const_get(namespaced_cop_class)
+      end
+
+      def safely_autocorrectable?(cop)
+        return true unless cop.cop_config
+        cop.cop_config.fetch("SafeAutoCorrect", true)
       end
     end
   end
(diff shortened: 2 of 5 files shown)


lib/standard/formatter.rb
```diff
@@ -15,6 +18,7 @@ def initialize(*args)
 
     def file_finished(file, offenses)
       uncorrected_offenses = offenses.reject(&:corrected?)
+      @autocorrectable_offenses = uncorrected_offenses.select(&method(:autocorrectable_offense?))
```
emilianoLeite: I used the (possibly) esoteric `method` method to avoid writing a 3-line block.

This line is equivalent to:
```ruby
@autocorrectable_offenses = uncorrected_offenses.select do |offense| 
  autocorrectable_offense?(offense)
end
```

I'm more than willing to make the aforementioned change if someone thinks that this kind of esoteric methods are best avoided.

test/standard/formatter_test.rb
```diff
@@ -1,7 +1,7 @@
 require "test_helper"
 
 class Standard::FormatterTest < UnitTest
-  Offense = Struct.new(:corrected?, :line, :real_column, :message)
+  Offense = Struct.new(:corrected?, :line, :real_column, :message, :cop_name)
```
emilianoLeite: I was really sad to have to change (almost) all the tests in the suite, but I actually don't know if it would be possible to implement this fix without doing so (though I can't really explain the reasoning behind this opinion).

test/standard/formatter_test.rb
```diff
@@ -26,7 +26,7 @@ def test_no_uncorrected_offenses_prints_nothing
 
   def test_does_not_print_fix_command_if_run_with_fix
     @subject = Standard::Formatter.new(@io, auto_correct: true, safe_auto_correct: true)
-    @subject.file_finished(@some_path, [Offense.new(false, 42, 13, "Neat")])
+    @subject.file_finished(@some_path, [Offense.new(false, 42, 13, "Neat", "Bundler/InsecureProtocolSource")])
```
emilianoLeite: Since we're leveraging all the `Cop` logic from Rubocop itself, I had to supply real `cop_name`s.

test/standardrb_test.rb
```diff
@@ -30,7 +29,6 @@ def test_project_with_cwd_in_nested_path
     refute status.success?
     assert_same_lines <<-MSG.gsub(/^ {6}/, ""), stdout
       standard: Use Ruby Standard Style (https://github.com/testdouble/standard)
-      standard: Run `standardrb --fix` to automatically fix some problems.
```
emilianoLeite: These lines were changed because this seemed the most "high-level" test of the suite and, therefore, should present the most authentic response possible.

danielmorrison:
You don't necessarily need to build up the list of `@autocorrectable_offenses`. You could just have a boolean and once set to true, you're done.

test/standard/formatter_test.rb
```diff
@@ -84,8 +84,8 @@ def test_prints_rake_message
   end
 
   def test_prints_call_for_feedback
-    @subject.file_finished(@some_path, [Offense.new(false, 42, 13, "Neat")])
-    @subject.file_finished(@some_path, [Offense.new(false, 43, 14, "Super")])
+    @subject.file_finished(@some_path, [Offense.new(false, 42, 13, "Neat", "Bundler/InsecureProtocolSource")])
+    @subject.file_finished(@some_path, [Offense.new(false, 43, 14, "Super", "Bundler/InsecureProtocolSource")])
```
searls: Would it be better to just address the root cause by doing whatever it wants us to do to Standard's Gemfile than to churn these tests in an incidental way?

lib/standard/helpers/autocorrectable.rb
```diff
@@ -0,0 +1,18 @@
+module Standard
+  module Helpers
+    module Autocorrectable
+      def autocorrectable_offense?(offense)
+        cop_instance(offense.cop_name).support_autocorrect?
```
searls: IIRC, `standardrb --fix` only runs what-rubocop-deems-to-be "_safe_" autocorrections. Could you verify that's the case?

If so, this check should probably also only return true for safe autocorrections

searls:
Thanks for submitting this! See my comments/questions and I'll look at merging

lib/standard/helpers/autocorrectable.rb
emilianoLeite: > IIRC, `standardrb --fix` only runs what-rubocop-deems-to-be "_safe_" autocorrections. Could you verify that's the case?

Yes, I believe that's exactly what [`MergesSettings`](https://github.com/testdouble/standard/blob/master/lib/standard/merges_settings.rb#L31) does

> If so, this check should probably also only return true for safe autocorrections

I partially reimplemented the [`RuboCop::Cop::AutocorrectLogic#autocorrect_enabled?`](https://github.com/rubocop-hq/rubocop/blob/master/lib/rubocop/cop/autocorrect_logic.rb#L19) method in f793685362529d1e41d6ccf84aadbbd3b61c8dd7. We could use the _actual_ implementation if our default value (i.e. when the `--fix` option is not supplied) for the `safe_auto_correct` flag is `true`, but I'm not sure if that's the case (I struggled to understand from where `LoadsYamlConfig` gets its default values). 

@searls , could you help me figure this out?

test/standard/formatter_test.rb
emilianoLeite: @searls I'm sorry, but I didn't fully understand what you meant. Could you please rephrase it?

From my point of view: these tests are getting churned because we're passing in a stubbed version of the `Offense` class. 
Since `Formatter#file_finished` now checks if the offense is autocorrectable (through the offense's `cop_name`), in the tests we must suply a stub which responds to the `cop_name` method.

I don't understand how doing anything to Standard's Gemfile would solve this issue.

searls:
Thanks for this! Landed in 0.0.29.

searls:
Whoops, this branch is fully merged in but I forgot to ensure a merge commit got made. See 88bd402882dd6a6ffca1f28d2d737b3d2a4885c2

Issue #35, opened 2018-12-01
Wrong whitespace correction


will:
Just noticed this auto change, which seems wrong. It may be a problem with rubocop though? If so let me know, and I'll open an issue there

![image](https://user-images.githubusercontent.com/1973/49331103-0cc00d80-f54d-11e8-9e3d-0ec565b5b1f1.png)

What's weird is that the rest of the table it didn't touch, just the last one.

searls:
If you'd asked me to guess, I would have thought standard would have eliminated all whitespace (beyond a single space) between the elements. Maybe I disabled that and now we're in a semi-poorly defined middle ground?

danielmorrison:
I have one repo where it collapses whitespace in one hash, but not another. I haven't taken the time to figure out why. I blame Rubocop.

danielmorrison:
Here's an example test case from said project: 

Before: 
```ruby
enum event_type: {
     thing_1:           0,
     thing_2:           1,
     longer_thing:      2,
     even_longer_thing: 3
}


config.fog_credentials = {
  provider:              'AWS',
  aws_access_key_id:     ENV['S3_ACCESS_KEY'],
  aws_secret_access_key: ENV['S3_SECRET'],
  region:                ENV['S3_REGION']
}
```

After `standardrb --fix`

```ruby
enum event_type: {
  thing_1: 0,
  thing_2: 1,
  longer_thing: 2,
  even_longer_thing: 3,
}

config.fog_credentials = {
  provider:              "AWS",
  aws_access_key_id:     ENV["S3_ACCESS_KEY"],
  aws_secret_access_key: ENV["S3_SECRET"],
  region:                ENV["S3_REGION"],
}
```

Interestingly, changing the second one to a simple variable assignment (`hash = {…`) _does_ collapse whitespace.

JoshCheek:
I haven't looked at the original example, but @danielmorrison's is because of line 27 in config/base.yml, which says to align hashes unless those hashes are the last argument to a method call.

https://github.com/testdouble/standard/blob/14da557e62e62e1a0f6639fd748dfc8731d83707/config/base.yml#L25-L27

The example that doesn't get aligned desugars to this:

```ruby
# the method is named "fog_credentials="
config.fog_credentials=({
  provider:              'AWS',
  aws_access_key_id:     ENV['S3_ACCESS_KEY'],
  aws_secret_access_key: ENV['S3_SECRET'],
  region:                ENV['S3_REGION']
})
```

[...]

PR #78, opened 2019-02-08
Passing a hash to a setter gets treated like data, not keyword args

JoshCheek:
Without this patch, hash literals passed to setters are treated like keyword args (ie the middle one should be treated like the top one, but it's instead treated like the bottom one:

image

With this patch, it gets treated like the top one:

image

---

* Note that I am not condoning the setting, I'm only fixing the inconsistency.
* Note that this implicitly implies that keyword options should not have an explicit hash wrapped around them (or they'll be seen as data, not an argument structure).
* For more context, see @danielmorrison's example [here](https://github.com/testdouble/standard/issues/35#issuecomment-461821683)

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -24,7 +24,7 @@ Layout/AlignArray:
 
 Layout/AlignHash:
   Enabled: true
-  EnforcedLastArgumentHashStyle: always_ignore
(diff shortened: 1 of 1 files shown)


searls:
Closed by 5b1ccfd

PR #97, opened 2019-03-14
Adds TD code of conduct to README.md

ericqweinstein:
Does what it says on the tin! All feedback welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,3 +352,5 @@ this to your `.vimrc`:
 let g:ale_fix_on_save = 1
 ```
 
+## Code of Conduct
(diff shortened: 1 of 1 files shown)


searls:
Propose we link to a single canonical CoC for all @testdouble projects and then link to it from the README, as opposed to having a separate policy copied/committed/maintained for every single repo. I own or help maintain between 300-500 repos, and keeping duplicated CoCs in sync across all of them wouldn't be maintainable.

ericqweinstein:
@searls Makes total sense. I'll touch base with @crebma and update shortly!

ericqweinstein:
Now with 100% less Markdown!

PR #108, opened 2019-04-13
undef before override, fixes #107

### LINKED ISSUE #107: Warning when required
Sidekiq is using Standard, thanks for this project Justin and team!

Since I run with warnings enabled, I see this when my test suite runs.

```
/Users/mikeperham/.gem/ruby/2.6.1/gems/parser-2.6.2.1/lib/parser/lexer.rb:10862: warning: assigned but unused variable - testEof
/Users/mikeperham/.gem/ruby/2.6.1/gems/standard-0.0.37/lib/standard/rubocop/ext.rb:3: warning: method redefined; discarding old message
/Users/mikeperham/.gem/ruby/2.6.1/gems/rubocop-0.67.2/lib/rubocop/cop/lint/assignment_in_condition.rb:51: warning: previous definition of message was here
```

Mayhaps one or more can be fixed easily?

mperham:
By undefining the method before redefining it, we avoid a runtime warning from Ruby which can pollute downstream test suites which run with warnings enabled.

### DIFF
--- a/lib/standard/rubocop/ext.rb
+++ b/lib/standard/rubocop/ext.rb
@@ -1,5 +1,6 @@
 module RuboCop
   class Cop::Lint::AssignmentInCondition
+    undef_method :message
     def message(_)
       "Wrap assignment in parentheses if intentional"
     end


searls:
Thanks Mike! I think our CI currently only runs against a recent Ruby like 2.5 or 2.6—any chance this would break on older Rubies? (We actually support all the way back to 1.8)

mperham:
It’s in 1.9.3 but I didn’t look back further.

ericqweinstein:
@searls @mperham It [does appear to exist](https://ruby-doc.org/core-1.8.7/Module.html#method-i-undef_method) in 1.8.7—I can try to install 1.8.5, 1.8.6, and 1.8.7 and test locally in the next day or two.

mperham:
👍 Thanks, Eric.  Is there a writeup of the rationale for supporting 1.8?  It's been EOL for six years now.

searls:
@mperham thanks for the reminder that we need to decide and document an actual policy with respect to support. In fact, I now realize I was wrong in my original reply to mention 1.8; I should have said 2.2. 

For instance, because Rubocop can parse Ruby versions earlier than the running ruby version, we consider both the runtime versions and target language versions.

* For runtimes, Standard supports 2.2 & up, the same as Rubocop. I imagine the supported runtime will continue to drop EOL'd Rubies as time goes on.
* For code analysis, Standard supports 1.8 & up. I'd like to keep this support as long as our dependencies can deal with it.

I wanted Standard to be able to be adopted by libraries that continue to support very old Rubies, and the work to do this was (and is) really minimal on our parts. Basically all, Standard does for older languages is define language-version-specific overrides in [its config directory](https://github.com/testdouble/standard/tree/master/config) in which it disables or adjusts rules that would lead to broken code if Standard were run against library code that supports older Rubies. (This is governed by the `ruby_version` in one's `.standard.yml`)

searls:
Landed in 0.0.38

PR #111, opened 2019-04-19
Remove the while until modifier rule.

samphippen:
This would force users to replace:

```ruby
while a == b
  a
end
```

with

```ruby
a while a == b
```

which is inconsistent and increases git churn.


searls:
Approved for two reasons:

* Pushing people towards one-liner loops is confusing and too easy to overlook when scanning. A potential infinite loop should probably have some indentation to call attention to itself
* This rule leads to more git churn when loops increase from 1 to >1 lines

PR #118, opened 2019-05-29
Fix semantic blocks when chaining a lonely operator

### LINKED ISSUE #109: Semantic blocks
I'm a long time user of the Weirich style and like that you don't enforce the line count thing on me. However, I'm not sure whether it's a good idea to attempt to detect whether a block is functional or not. I ran this on our project and got several things it got wrong:

```ruby
CONCENTRATION.find { |end_date, _threshold|
  end_date >= Time.zone.today
}&.last || 0
```
was changed to the `do ... end`

```ruby
@_document ||= Prawn::Document.new do
  font_families.update(
    "Courier" => {
      normal: "app/assets/fonts/source-code-pro/SourceCodePro-Regular.ttf",
      bold: "app/assets/fonts/source-code-pro/SourceCodePro-Bold.ttf"
    }
  )
end
```
was changed to `{ ... }`

The Weirich style is imnsho close to impossible to enforce correctly (unless you leave many cases untouched, which makes it kinda useless) and isn't easy to understand for newer developers.

The line count style is just useless. The only advantage is it's wide usage.

We should decide, whether our affection for Jim Weirich and our wish to appear smart is worth the costs of this complication. If I'm honest to myself, I can't remember a case when I actually used this signal whether the block is functional or procedual.

To me it's similar to the double vs. single quotes thing for strings. Maybe it's even easier, because you don't need to care about the content of the block, while with strings, you might want to change the style depending whether you have quotes in it.

My preferences in order:
1) Always enforce braces
2) Don't enforce the block delimiters at all, so we can use the Weirich style
3) The Weirich style enforcement attempt or the line count style (With a bad feeling in my stomach)

searls:
- [ ] make the test pass

Fixes #109

### DIFF
--- a/test/standard/cop/semantic_blocks_test.rb
+++ b/test/standard/cop/semantic_blocks_test.rb
@@ -57,4 +57,12 @@ class RuboCop::Cop::Standard::SemanticBlocksTest < UnitTest
       end
     RUBY
   end
+
+  def test_lonely_chains_are_functional
+    assert_no_offense @cop, <<-RUBY
+      CONCENTRATION.find { |end_date, _threshold|
+        end_date >= Time.zone.today
+      }&.last || 0
+    RUBY
+  end
 end


Yunus259 (approved):

searls:
Obsolete!

Issue #119, opened 2019-06-21
Compatibility with ruby 2.7


hurricup:
Attempt to run standardrb on ruby-27 fails, probably because of rubocop's version. Would be nice to address this.

searls:
I think this will involve implementing #105 by upgrading rubocop, pulling in rubocop-performance, and making sure everything passes

searls:
Thanks @koic!

JamesChevalier:
I think this was resolved by https://github.com/testdouble/standard/pull/126

searls:
Yup!

PR #120, opened 2019-06-26
Support Ruby 2.7

yez:
Addresses #119

With this version of Standard, we will be dropping support of Ruby <=
2.3 based on dependencies.

Rubocop does not support Ruby 2.7 until [this commit](https://github.com/rubocop-hq/rubocop/commit/c06cc51c541c3ac11ab529c909a8a0f90eefa892)
In that commit, the rubocop gemspec file requires Ruby >= 2.3.

We have also pulled in [rubocop-performance](https://github.com/rubocop-hq/rubocop-performance) instead of specifying performance cops in `config/base.yml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -279,8 +279,8 @@ for details.
 
 Because Standard wraps RuboCop, they share the same [runtime
 requirements](https://github.com/rubocop-hq/rubocop#compatibility)—currently,
-that's MRI 2.2 and newer. While Standard can't avoid this runtime requirement,
-it does allow you to lint codebases that target Ruby versions older than 2.2 by
+that's MRI 2.3 and newer. While Standard can't avoid this runtime requirement,
+it does allow you to lint codebases that target Ruby versions older than 2.3 by
 narrowing the ruleset somewhat.
 
 Standard will default to telling RuboCop to target the currently running version
@@ -288,7 +288,7 @@ of Ruby (by inspecting `RUBY_VERSION` at runtime. But if you want to lock it
 down, you can specify `ruby_version` in `.standard.yml`.
 
 ```
-ruby_version: 1.8.7
+ruby_version: 2.3.0
 ```
 
 See
@@ -297,11 +297,11 @@ for an example.
 
 It's a little confusing to consider, but the targeted Ruby version for linting
 may or may not match the version of the runtime (suppose you're on Ruby 2.5.1,
-but your library supports Ruby 2.2.0). In this case, specify `ruby_version` and
+but your library supports Ruby 2.3.0). In this case, specify `ruby_version` and
 you should be okay. However, note that if you target a _newer_ Ruby version than
 the runtime, RuboCop may behave in surprising or inconsistent ways.
 
-If you are targeting a Ruby older than 2.2 and run into an issue, check out
+If you are targeting a Ruby older than 2.3 and run into an issue, check out
 Standard's [version-specific RuboCop
 configurations](https://github.com/testdouble/standard/tree/master/config) and
 consider helping out by submitting a pull request if you find a rule that won't
--- a/config/base.yml
+++ b/config/base.yml
@@ -1,5 +1,7 @@
+require: rubocop-performance
+
 AllCops:
-  # Prevent RuboCop from exploding when it finds an older-than-2.2 .ruby-version
+  # Prevent RuboCop from exploding when it finds an older-than-2.3 .ruby-version
   TargetRubyVersion: 2.5
   DisabledByDefault: true
   Exclude: []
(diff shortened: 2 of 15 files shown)


README.md
```diff
@@ -279,16 +279,16 @@ for details.
...
+that's MRI 2.3 and newer. While Standard can't avoid this runtime requirement,
+it does allow you to lint codebases that target Ruby versions older than 2.3 by
 narrowing the ruleset somewhat.
 
 Standard will default to telling RuboCop to target the currently running version
 of Ruby (by inspecting `RUBY_VERSION` at runtime. But if you want to lock it
 down, you can specify `ruby_version` in `.standard.yml`.
 
 ```
-ruby_version: 1.8.7
```
searls: Is there a reason you changed this line? You should still be able to lint 1.8.7 code (while executing rubocop under 2.3), right?

config/base.yml
```diff
@@ -474,63 +476,6 @@ Naming/UncommunicativeBlockParamName:
...
-
-Performance/StringReplacement:
-  Enabled: true
-
-Performance/UnfreezeString:
-  Enabled: true
-
-Performance/UriDefaultParser:
-  Enabled: true
-
```
searls: I expected these to still be turned on -- can we leave them on?

Subtext: I suspect this might have been deleted if rubocop-performance is enabled-by-default for all its cops with something like `AllCops/DisabledByDefault: false`, and that we should find that, turn it off, and then go back to expressly enabling them. 

context: With standard we're conscientious about disabling-by-default and expressly enabling cops so that underlying rubocop{-performance} upgrades don't break us as often.

config/ruby-1.8.yml
```diff
@@ -1,8 +0,0 @@
-inherit_from: ./ruby-1.9.yml
```
searls: Apologies for being dense—what necessitated the pre-2.2 configs being deleted? My understanding was rubocop had to be _run_ under 2.2 or later, but the way standard works is that while it must _execute_ 2.2, it can detect and lint 1.8 _code listings_

test/fixture/runner/agreeable.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
searls: Was this necessary for something?

config/base.yml
yez: We're requiring `rubocop-performance` in this file. Is your thought that some of these cops _within_ `rubocop-performance` are not enabled?

config/ruby-1.8.yml
yez: I might be missing something crucial here. Are you saying that this code needs to be able to lint code outside of its own runtime? i.e. I run `rubocop path/to/some/dir` and inside `path/to/some/dir` there is ruby 1.8 code but from my working directory where I ran rubocop, it uses ruby >= 2.3? 

We still need whatever is running this `standard` to be >= Ruby 2.3 due to the commit I linked in this PR description.

test/fixture/runner/agreeable.rb
yez: Cops for Ruby >= 2.3 require this frozen string literal comment. Since I thought we were bumping `standard`'s min ruby version to 2.3, adding this for tests made sense to me.

config/base.yml
searls: Sorry, my point is that we have set up rubocop to explicitly enable each cop. We don't leave them all turned on by default—I'm sure we'd want to continue doing that with rubocop-performance's cops as well

config/ruby-1.8.yml
searls: Yes, exactly. Running rubocop under MRI 2.2, standard can lint codebases that support 1.8, that's what all these special one-off configs are here for.

test/fixture/runner/agreeable.rb
searls: Let's definitely disable whatever cop is enforcing this — one possibility based on the above comment about performance being on-by-default, this got re-turned-on when it was intentionally turned off earlier

config/base.yml
yez: Understood, will change.

config/ruby-1.8.yml
yez: Ah, that makes a lot more sense. I didn't think of that use case, thanks!

test/fixture/runner/agreeable.rb
yez: Ok sounds good!

yez: I'm having an issue making this fixture and it's related test pass. The test that is failing is: 

```
  1) Failure:
Standard::Runners::RubocopTest#test_empty_output_on_quiet_success [/Users/jakeyesbeck/standard/test/standard/runners/rubocop_test.rb:20]:
--- expected
+++ actual
@@ -1 +1,8 @@
-""
+"== test/fixture/runner/agreeable.rb ==
+" +
+"C:  1:  1: Style/FrozenStringLiteralComment: Missing magic comment # frozen_string_literal: true.
+" +
+"
+" +
+"1 file inspected, 1 offense detected
+"
```

The test fails on the frozen string cop despite me adding to `config/base.yml`:

```
Style/FrozenStringLiteralComment:
  Enabled: false
```

@searls any ideas?

searls:
Hey @yez, could you take a look at https://github.com/yez/standard/pull/1 and see about merging so we can get back to green?

searls:
Closing in favor of #126

PR #121, opened 2019-06-27
enforce %x(style invocation) because backticks `` are deprecated

searls:
Not a lot of people know/realize this, but Matz & ruby-core are deprecating backticks for running system commands and planning to eliminate them in Ruby 3 to free up the ` character for future syntax features. ([RedMine issue](https://bugs.ruby-lang.org/issues/14385))

As a result I think it's prudent that standard start enforcing the future-proof shorthand for invoking system commands

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -737,7 +737,7 @@ Style/ColonMethodDefinition:
 
 Style/CommandLiteral:
   Enabled: true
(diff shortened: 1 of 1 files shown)


searls:
Ended up releasing 0.0.41 which allows either style, thanks to @stevenjackson's advice

Issue #122, opened 2019-06-27
After 1.0, will standard continue to update alongside rubocop?


onebree:
I understand that standard is set up to opt-into cops. While working on a PR related to cop settings, I noticed a few things:

There are plenty of cops omitted from `config/base.yml`, but without an attached issue/commit discussing _why_. For example:

  + All things metrics
  + Naming/PredicateName
  + Naming/RescuedExceptionsVariableName
  + Performance/RedundantBlockCall

There are already a couple of cops added, or renamed, since rubocop 0.67:

  + 0.68: Rename `Layout/FirstParameterIndentation` to `Layout/IndentFirstArgument`
    * This one in particular tripped me up while working on my PR. I was referencing the latest rubocop docs page, rather than the config file at version 0.67
  + 0.69 - Add new cop `Style/NegativeUnless` that checks for unless with negative condition

So, given that rubocop is also pre-1.0, how will standard grow alongside it, if at all? I guess it's unclear whether cops were opted out of by intention, or by accident. And I think clearing that up somewhere (maybe a running list of opted-in and opted-out docs?) would help people researching this as an alternative to custom rubocop configs, or even contributors like myself trying to understand the project more.

Thank you!

ktopolski:
Metrics are omitted on purpose, please see #102 . The plan is to grow along the rubyfmt which will take care of metrics IIRC.

I'm not so sure about other cops tho, probably we need to wait on @searls input here.

Cheers guys

searls:
Hi Hunter, thanks for opening this issue.

To be honest, I think a lot of the sense of informality that you're describing is a symptom of the fact that Standard is a linting tool _from a group of people who don't care too much about style guides_. Because the primary goal Standard sets out to solve is not "establish One True Style", but just to avert bikeshedding on teams, the answer to 80% of the configurations we've landed on are with one of these rationales:

• Most people don't care or it doesn't matter (e.g. Arbitrarily picking one `Enumerable` method over an alias method)
• It can't be applied reasonably in 100% of cases, so it's better to turn it off than to enforce it and require people to figure out how to work around it (e.g. every single `Metrics/*` cop, AFAIK)
• There's an obviously "correct" choice, because the cop pushes people away from a deprecation, performance risk, unsafe code execution, likely mistake, etc. (e.g. [Lint/SafeNavigationChain](https://rubocop.readthedocs.io/en/stable/cops_lint/#lintsafenavigationchain))

[...]

PR #123, opened 2019-06-30
Simple dockerized dev environment setup

ktopolski:
I had to set it up to work on #122 , maybe someone else could use it as well :)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,11 @@
+FROM ruby:2.5
+
+RUN mkdir /gem
+WORKDIR /gem
+
+COPY Gemfile Gemfile.lock standard.gemspec ./
+COPY lib/standard/version.rb ./lib/standard/
+
+RUN bundle install --jobs 8
+
+ADD . /gem/
--- a/README.md
+++ b/README.md
@@ -366,5 +366,17 @@ this to your `.vimrc`:
 let g:ale_fix_on_save = 1
 ```
 
+## Contributing
+
+To setup your dev environment you can use the Docker:
+
+```bash
+$ git clone https://github.com/testdouble/standard
+$ cd standard
+$ docker build . -it standard
+$ docker run -v `pwd`:/gem -it standard bash
+$ bundle exec rake
+```
+
 ## Code of Conduct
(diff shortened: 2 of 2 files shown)


larquin:
I have a few concerns about introducing a containerized environment in the standard project, and I'm not totally convinced it reduces the barrier to entry for contributing to standard. If anything, my thoughts are that it adds complexity to the project unnecessarily, as it requires us to maintain our code and the container itself. 

Containerized Environments assume a specific workflow, and can obfuscate troubleshooting issues. I'd advise us to avoid endorsing a specific workflow when we're working on standard. For developers unfamiliar with working in container based dev environments, merely wrestling with the state of the container, e.g. "is it up or down?", can cause unnecessary frustration. 

The image we're using in this PR also runs as root, which is something I wouldn't advise as a good standard of practice in any containerized environment. Ideally our container would set up a user, or change the user at container run time using an entry point script. 

If standard was an application, and we were building it for a specific containerized deployment, I'd be more open to containerizing our dev process, but if we're looking to add a container to alleviate development concerns in a library, I'm not certain we're addressing the pain points of working with different versions of say, rubocop.

ktopolski:
Hello Micah,

Thank you for your feedback! I always enjoy a constructive one.

I have no hard feelings for this one in particular. It's not like I'm trying to push a specific standard workflow here, it's more like there isn't any - neither in `README` nor in `CONTRIBUTING`. I wouldn't mind changing this PR to just include a "Contributing" section with few instructions on how to setup local env for standard development.

As for the Dockerfile itself, I was sacrificing good partice of non-root user set via entrypoint for easier readability. I do, of course, agree that it is best practice - I was just worried about the same thing you are in this comment - barrier of entry.

Given all of that, I guess we can either drop the docker based env for dedicated contributing section in readme or support both local and docker. Then, we could bring back best practices as the docker flow wouldn't be the only one mentioned in contributing, tchus people not familiar with it wouldn't hit the wall.

larquin:
Gotcha, I think we're aligned. 

Maybe we can drop the docker based environment and shore up our contributing section so people have a better handle on how to get started with standard. I definitely think we're lacking in that regard so it can be confusing on how to start hacking on the project.

ktopolski:
Closing this in favor of #124 , thanks for your input @larquin :)

PR #125, opened 2019-07-08
Enforce RuboCop defaults as needed

onebree:
Reference issue: #122 

CC @ktopolski for helping with this!

With the current Standard config, RuboCop defaults are assumed. Defaults from the supported RuboCop version were added where appropriate, as a means of self-documentation.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -6,15 +6,27 @@ AllCops:
 
 Bundler/DuplicatedGem:
   Enabled: true
+  Include:
+    - '**/*.gemfile'
+    - '**/Gemfile'
+    - '**/gems.rb'
 
 Bundler/InsecureProtocolSource:
   Enabled: true
+  Include:
+    - '**/*.gemfile'
+    - '**/Gemfile'
+    - '**/gems.rb'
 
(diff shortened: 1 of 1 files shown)


searls:
Thank you!

So that I and other readers understand: what this PR does is simply fix in place whatever Standard has already been doing, because it's been inheriting these defaults from Rubocop. This will serve as a guard against future changes to Rubocop defaults (for example, switching from symbols to hashrockets as the `EnforcedStyle` in hash literals) having hte adverse effect of automatically/immediately changing `standardrb`'s behavior without our knowledge/say-so.

Do I have that right?

onebree:
You're correct, @searls!

config/base.yml
```diff
@@ -738,12 +916,16 @@ Style/ColonMethodDefinition:
 Style/CommandLiteral:
   Enabled: true
   EnforcedStyle: mixed
```
onebree: @ktopolski I just realized you changed this to `backticks`. But I changed it to `mixed` to match the latest version of master (not reflected on my fork).

ktopolski: nice catch, thank you Hunter :)

PR #128, opened 2019-08-08
Set rubocop setting to disable extra spacing for alignment

### LINKED ISSUE #127: Assignations rule
I'm not sure if rubocop recommends or enforces any type of rule for assignations.
I have seen that some people like to align assignations on the equals operator like:

```ruby
foo       = 1
bar       = 2
foobar    = 3
foobarbas = 4
``` 

IMHO, I think standardrb should just default to not aligning on the equals (like the golang formatter does)
```ruby
foo = 1
bar = 2
fopbaar = 3
foobarbas = 4
```

I don't have strong arguments against the 1st option other than sometimes looks really ugly (subjective), it's width inefficient and you have to remember to update the other assignations (this could be solved by the formatter).

pablocrivella:
Fixes #127

Related links:

https://docs.rubocop.org/en/stable/cops_layout/#layoutextraspacing
https://github.com/rubocop-hq/rubocop/issues/7269

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -137,7 +137,7 @@ Layout/EndOfLine:
 
 Layout/ExtraSpacing:
   Enabled: true
-  AllowForAlignment: true
+  AllowForAlignment: false
   AllowBeforeTrailingComments: false
   ForceEqualSignAlignment: false
 


searls:
Landed in 0.1.2 -- thanks!

kimroen:
This could be intended, but this change also disallows spacing to align comments and in between arguments for alignment.

So this example Devise migration changed from this

```rb
class DeviseCreateAdminUsers < ActiveRecord::Migration[5.0]
  def change
    create_table :admin_users do |t|
      ## Database authenticatable
      t.string :email,              null: false, default: ""
      t.string :encrypted_password, null: false, default: ""

      ## Recoverable
      t.string   :reset_password_token
      t.datetime :reset_password_sent_at
    end

    add_index :admin_users, :email,                unique: true
    add_index :admin_users, :reset_password_token, unique: true
    add_index :admin_users, :confirmation_token,   unique: true
    add_index :admin_users, :unlock_token,         unique: true
  end
end
```

to this:

```rb
class DeviseCreateAdminUsers < ActiveRecord::Migration[5.0]
  def change
    create_table :admin_users do |t|
      ## Database authenticatable
      t.string :email, null: false, default: ""
      t.string :encrypted_password, null: false, default: ""

      ## Recoverable
      t.string :reset_password_token
      t.datetime :reset_password_sent_at
    end

    add_index :admin_users, :email, unique: true
    add_index :admin_users, :reset_password_token, unique: true
    add_index :admin_users, :confirmation_token, unique: true
    add_index :admin_users, :unlock_token, unique: true
  end
end
```

And this example of documenting configuration options changed from this

```rb
  config.amount_column = {prefix: "",           # column name prefix
                          postfix: "_cents",    # column name  postfix
                          column_name: nil,     # full column name (overrides prefix, postfix and accessor name)
                          type: :integer,       # column type
                          present: true,        # column will be created
                          null: false,          # other options will be treated as column options
                          default: 0,}
```

to this

```rb
  config.amount_column = {prefix: "", # column name prefix
                          postfix: "_cents", # column name  postfix
                          column_name: nil, # full column name (overrides prefix, postfix and accessor name)
                          type: :integer, # column type
                          present: true, # column will be created
                          null: false, # other options will be treated as column options
                          default: 0,}
```

It also caught some places where we had accidentally added extra spaces in between arguments, so I'm not necessarily arguing that we shouldn't have this rule. I just wanted to see that it was done on purpose.

We could use these arguments to tweak any of these: https://docs.rubocop.org/en/stable/cops_layout/#configurable-attributes_17

pablocrivella:
IMHO I think most of the times everyone forgets to update the alignment, so the practical approach would be not to align anything 🤷‍♂

PR #141, opened 2019-10-12
Ignore the /vendor directory.

Insti:
When the support gems are installed in the `/vendor` directory[1] the gems are included in the test coverage report output to `coverage/index.html`

We want the test coverage report to only include the `standard` code.

This PR adds an exclusion to the simplecov configuration so that the external gem code is not included in the code coverage report.

It also adds `/vendor/` to the `.gitignore` file.

Note: It could be argued that this change does not belong in the main repository and this could be handled entirely by local branch configuration, in which case just close this PR.

[1]  Using `$ bundle install --path vendor/bundle` (see: https://bundler.io/bundle_install.html)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /pkg/
 /spec/reports/
 /tmp/
+/vendor/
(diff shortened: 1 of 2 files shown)


searls:
LGTM Thanks!

PR #145, opened 2019-10-28
Add tests for single-line block exception

Insti:
In https://github.com/testdouble/standard/pull/9 the decision was made that single-line blocks should always be an exception to the semantic blocks rule.

This commit adds explicit tests for single line cases in an attempt to document this exception.


searls:
Thanks!

PR #155, opened 2019-12-18
Added the Ability to Generate a todo File

mrbiggred:
Useful if you have an existing project that you want to incorporate Standard into. Run the below to generate the .standard_todo.yml file:

```
standardrb --gen-ignore
```

I thought it would be better to have a separate ignore file. Easier to generate. Also allows developers to separate the files that are "temporarily" ignored versus the files to permanently ignore in the `.standard.yml`.

Looking forward to your feedback.  Thank you

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,7 +90,7 @@ If you have an existing project but aren't ready to fix all the files yet you ca
 generate a todo file:
 
 ```bash
-$ bundle exec standardrb --gen-ignore
+$ bundle exec standardrb --generate-todo
 ```
 
 This will create a `.standard_todo.yml` that lists all the files that contain errors.
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ cannot be found by ascending the current working directory (i.e. against a
 temporary file buffer in your editor), you can specify the config location with
 `--config path/to/.standard.yml`.
 
-Similar with the `.standard_todo.yml` you can specify `--todo path/to/.standard_todo.yml`.
+Similarly, fow the `.standard_todo.yml` file, you can specify `--todo path/to/.standard_todo.yml`.
 
 ## What you might do if you're REALLY clever
 
--- a/README.md
+++ b/README.md
@@ -54,7 +54,9 @@ flag.
 
 If you're familiar with [RuboCop](https://github.com/rubocop-hq/rubocop), you
 can look at Standard's current base configuration in
-[config/base.yml](/config/base.yml).
+[config/base.yml](/config/base.yml). In lieu of a separate changelog file,
+significant changes to the configuration will be documented as [GitHub release
+notes](https://github.com/testdouble/standard/releases).
 
 **[NOTE: until StandardRB hits 1.0.0, we consider this configuration to be a
 non-final work in progress and we encourage you to submit your opinions (and
@@ -151,7 +153,7 @@ ignore:                 # default: []
 ```
 
 Note: If you're running Standard in a context where your `.standard.yml` file
-cannot be found by ascending the current working directory (i.e. against a
+cannot be found by ascending the current working directory (i.e., against a
 temporary file buffer in your editor), you can specify the config location with
 `--config path/to/.standard.yml`.
 
@@ -206,6 +208,9 @@ if you've used StandardJS.)
 * [Hashrocket](https://hashrocket.com)
 * [Brand New Box](https://brandnewbox.com)
 * [Monterail](https://www.monterail.com)
+* [Level UP Solutions](https://levups.com)
+* [Honeybadger](https://www.honeybadger.io)
+* [Amazon Web Services](https://aws.amazon.com/)
 * And that's about it so far!
 
 If your team starts using Standard, [send a pull
@@ -236,7 +241,7 @@ No. The whole point of Standard is to save you time by avoiding
 about code style. There are lots of debates online about tabs vs. spaces, etc.
 that will never be resolved. These debates just distract from getting stuff
 done. At the end of the day you have to 'just pick something', and that's the
-whole philosophy of Standard -- its a bunch of sensible 'just pick something'
+whole philosophy of Standard -- it's a bunch of sensible 'just pick something'
 opinions. Hopefully, users see the value in that over defending their own
 opinions.
 
@@ -366,11 +371,12 @@ feedback loop. Some editors support asynchronously running linters.
 
 ## Does Standard work with [Insert other tool name here]?
 
-Maybe! Start by searching the repository to see if there's an existing issue open for 
+Maybe! Start by searching the repository to see if there's an existing issue open for
 the tool you're interested in. That aside, here are other known integrations aside
 from editor plugins:
 
 * [Pronto](https://github.com/julianrubisch/pronto-standardrb)
+* [Spring](https://github.com/lakim/spring-commands-standard)
 
 ## Contributing
 
@@ -385,4 +391,10 @@ $ bundle exec rake # to run test suite
 ```
 
 ## Code of Conduct
-This project follows Test Double's [code of conduct](https://testdouble.com/code-of-conduct) for all community interactions, including (but not limited to) one-on-one communications, public posts/comments, code reviews, pull requests, and GitHub issues. If violations occur, Test Double will take any action they deem appropriate for the infraction, up to and including blocking a user from the organization's repositories.
+
+This project follows Test Double's [code of
+conduct](https://testdouble.com/code-of-conduct) for all community interactions,
+including (but not limited to) one-on-one communications, public posts/comments,
+code reviews, pull requests, and GitHub issues. If violations occur, Test Double
+will take any action they deem appropriate for the infraction, up to and
+including blocking a user from the organization's repositories.
--- a/README.md
+++ b/README.md
@@ -86,6 +86,20 @@ $ bundle exec standardrb "lib/**/*.rb" test
 **Note:** by default, StandardRB will look for all `*.rb` files (and some other
 files typically associated with Ruby like `*.gemspec` and `Gemfile`)
 
+If you have an existing project but aren't ready to fix all the files yet you can
+generate a todo file:
+
+```bash
+$ bundle exec standardrb --gen-ignore
+```
+
+This will create a `.standard_todo.yml` that lists all the files that contain errors.
+When you run Standard in the future it will ignore these files as if they lived under the
+`ignore` section in the `.standard.yml` file.
+
+As you refactor your existing project you can remove files from the list.  You can
+also regenerate the todo file at anytime by re-running the above command.
+
 ### Using with Rake
 
 Standard also ships with Rake tasks. If you're using Rails, these should
@@ -140,6 +154,8 @@ Note: If you're running Standard in a context where your `.standard.yml` file
 cannot be found by ascending the current working directory (i.e. against a
 temporary file buffer in your editor), you can specify the config location with
 `--config path/to/.standard.yml`.
+
+Similar with the `.standard_todo.yml` you can specify `--todo path/to/.standard_todo.yml`.
 
 ## What you might do if you're REALLY clever
 
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ cannot be found by ascending the current working directory (i.e. against a
 temporary file buffer in your editor), you can specify the config location with
 `--config path/to/.standard.yml`.
 
-Similarly, fow the `.standard_todo.yml` file, you can specify `--todo path/to/.standard_todo.yml`.
+Similarly, for the `.standard_todo.yml` file, you can specify `--todo path/to/.standard_todo.yml`.
 
 ## What you might do if you're REALLY clever
 
--- a/lib/standard/loads_yaml_config.rb
+++ b/lib/standard/loads_yaml_config.rb
@@ -13,18 +13,15 @@ module Standard
       yaml_path = @parses_cli_option.call(argv, "--config") ||
         FileFinder.new.call(".standard.yml", search_path)
 
-      # REM
+      # Check for the todo file.  If found merge with standard config file.
       todo_yaml_path = @parses_cli_option.call(argv, "--todo") ||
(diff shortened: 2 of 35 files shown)


mrbiggred:
It appears the "cleanpath" error is a known Ruby issue:

https://bugs.ruby-lang.org/issues/10011

I didn't encounter this issue as I used Ruby 2.6 when testing.  I'll update my code so it works on Ruby 2.4.1, which is what the tests appear to run on.

The second error with the path names not lining up is interesting.  I think something is bleeding over between the the new `genignore_test.rb` tests I added and the existing `rubocop_test.rb` tests.

mrbiggred:
Sorry but I won't be able to fix the above issues till after the holidays.  Let me know having a PR open for that long will cause any issues and we can figure something out, like closing this one and re-submitting a new one in the new year.

Thanks and Happy Holidays!

mrbiggred:
The error with Ruby 2.4.1 and `relative_path_from` is fixed.  Still can't figure out why the path is incorrect if my new runner tests runs before the existing Rubocop runner test but I'll keep poking around till I find the problem.

searls:
Thanks for your continued efforts!

mrbiggred:
Just a heads up that I plan on taking another look at the failing tests this Thursday.  It appears that the underlying Rubocop settings are persisted between the tests.  At least that is my current theory.  Does anyone has any suggestions on how to clear the Rubocop settings between tests?  Thanks.

mrbiggred:
I have finally found and fixed the failing test issue.  The problem was Rubocop was caching the working directory.  This combined with how simple formatter works would cause the `test_print_corrected_output_on_stdin` in `test/standard/runners/rubocop_test` to fail if it ran before the new `test_todo_generated` test I added in `test/standard/runners/genignore_test`.

Note: the quiet formatter inherits from simple formatter so will have the same problem.

To fix the problem I added a Rubcop path reset to `test/test_helper`:

```ruby
def setup
  RuboCop::PathUtil.reset_pwd
end
```

I then changed the `test/standard/runners/rubocop_test` and `test/standard/runners/genignore_test` setup methods to use the base setup:

```ruby
def setup
  super

  @subject = Standard::Runners::Genignore.new
end
```

Let me know if you can think of a better was to fix this failing test issue.


**Debugging Notes**
The code where Rubocop caches the current working directory can be found below:

https://github.com/rubocop-hq/rubocop/blob/31fe1a58d196f474991daf64d97fe64e9836bb71/lib/rubocop/path_util.rb#L57

This is the commit that added the caching to Rubocop:

https://github.com/rubocop-hq/rubocop/pull/5137/files

A description of how simple formatter with either display the relative or absolute path can be found below.  I couldn't find anywhere on the main Rubocop documentation that describes how simple formatter decides the if it should use relative or absolute paths.

https://github.com/rubocop-hq/rubocop/pull/262#issuecomment-19265766

The simple formatter code uses the `smart_path` method to output the file name.  The `smart_path` method will use the cached working directory:

https://github.com/rubocop-hq/rubocop/blob/31fe1a58d196f474991daf64d97fe64e9836bb71/lib/rubocop/formatter/simple_text_formatter.rb#L41

README.md
```diff
@@ -141,6 +155,8 @@ cannot be found by ascending the current working directory (i.e. against a
 temporary file buffer in your editor), you can specify the config location with
 `--config path/to/.standard.yml`.
 
+Similarly, fow the `.standard_todo.yml` file, you can specify `--todo path/to/.standard_todo.yml`.
```
searls:
```suggestion
Similarly, for the `.standard_todo.yml` file, you can specify `--todo path/to/.standard_todo.yml`.
```

lib/standard/loads_yaml_config.rb
```diff
@@ -12,11 +12,25 @@ def initialize
...
         FileFinder.new.call(".standard.yml", search_path)
-      construct_config(yaml_path, load_standard_yaml(yaml_path))
+      standard_yaml = load_standard_yaml(yaml_path)
+
+      todo_yaml_path = @parses_cli_option.call(argv, "--todo") ||
+        FileFinder.new.call(".standard_todo.yml", search_path)
+      todo_yaml = load_standard_yaml(todo_yaml_path)
+
+      merge_ignore(standard_yaml, todo_yaml)
+      construct_config(yaml_path, standard_yaml)
```
searls: I'd at least reorder the body to put the two yaml files next to each other, but better yet since they do the same thing, extract that to a method, so that the duplication is gone and all 3 things the method does are at the same level of abstraction

lib/standard/merges_settings.rb
```diff
@@ -41,8 +41,11 @@ def determine_command(argv)
         :help
       elsif (argv & ["--version", "-v"]).any?
         :version
+      elsif (argv & ["--gen-ignore"]).any?
```
searls: Is `--gen-ignore` the exact arg name rubocop uses? It's not clear at all to me what this flag actually does if I'm not literally thinking about this feature. I'd rather it was named something longer and intent-revealing `--generate-ignore-todo-file`

README.md
mrbiggred: Fixed.  Thanks for spotting my type.

lib/standard/loads_yaml_config.rb
mrbiggred: Refactored to what I think you are asking for.  Let me know if I misunderstood.

lib/standard/merges_settings.rb
mrbiggred: How does `--generate-todo` sound?

mrbiggred:
@searls thank you for your feedback.  I've made the changes you suggested.  Let me know what you think.

README.md
```diff
@@ -90,7 +90,7 @@ If you have an existing project but aren't ready to fix all the files yet you ca
 generate a todo file:
 
 ```bash
-$ bundle exec standardrb --gen-ignore
```
searls: Ignorant comment for still having not pulled down the code to run it:

Does standard print a warning when it's running in "todo" mode? I'm ok with naming it "todo" only if that's the case. Come to think of it, printing a progress-to-completion whenever a todo ignore was present would be pretty great.

What I want to avoid is anyone using this as a crutch to think that they're "passing" standard when the presence of a file is actually giving them a free pass

mrbiggred: @searls if I understand the above correctly you would like Standard to notify the user that the todo file is being used.  Currently it does not but I like that idea.

Something like:

```
$ bundle exec standardrb

Todo file found, ignoring the following files:
- file_a.rb
- file_b.rb
```

Let me know if I misunderstood what you are asking for.  Thanks.

searls: Yes, something like this would be good. I might make it slightly more scary:

```
WARNING: this project is being migrated to standard gradually via `whatevertodo.yml` and is ignoring these files:
```

searls:
Howdy @mrbiggred -- do you have what you need from me to get this over the finish line, you think?

mrbiggred:
@searls thank you for reminding me about this pull request.  With all the coronavirus craziness I forgot about it.  I took a look this morning and think I see where I would add the message and am working on getting a list of the files ignored.  I'll try to continue my work later this week but if you don't hear from me by next week don't be afraid to ping me again.

mrbiggred:
@searls I have added the warning message if the todo file is being used.  I struggled to figure out the best way to store and print out the files being ignored.  If you notice a better way please let me know.

The merge conflicts are a result of standard warning me about missing trailing commas.

I will try this branch on an actual product after the long weekend for some beta testing.

lib/standard/runners/genignore.rb
```diff
@@ -0,0 +1,43 @@
...
+module Standard
+  module Runners
+    class Genignore
+      def call(config)
+        # Run Rubocop to generate the files with errors.
+        config.rubocop_options[:formatters] = [["files", "temp_exclude.txt"]]
+        config.rubocop_options[:format] = "files"
+        config.rubocop_options[:out] = "temp_exclude.txt"
+
+        File.delete("exclude.txt") if File.exist?("temp_exclude.txt")
```
searls, line 13: What's `exclude.txt`? It's only mentioned here.

As for `temp_exclude.txt`, I don't love spitting out a temporary file into a person's project directory, because if there's a failure it'll sit there and they won't know what it is. Maybe it'd be better to use Ruby's [tempfile](https://ruby-doc.org/stdlib-2.3.0/libdoc/tempfile/rdoc/Tempfile.html) instead?

lib/standard/loads_yaml_config.rb
```diff
@@ -25,15 +22,17 @@ def load_standard_yaml(yaml_path)
...
 
-    def construct_config(yaml_path, standard_yaml)
+    def construct_config(yaml_path, standard_yaml, todo_path, todo_yaml)
       {
         ruby_version: Gem::Version.new((standard_yaml["ruby_version"] || RUBY_VERSION)),
         fix: !!standard_yaml["fix"],
         format: standard_yaml["format"],
         parallel: !!standard_yaml["parallel"],
-        ignore: expand_ignore_config(standard_yaml["ignore"]),
+        ignore: expand_ignore_config(standard_yaml["ignore"]) + expand_ignore_config(todo_yaml["ignore"]),
```
searls, line 31: What happens if there are duplicates or contradictions between these two sets of ignore rules?

lib/standard/runners/genignore.rb
mrbiggred: I've updated to the code to use Ruby's tempfile.  Thanks for the tip.

lib/standard/loads_yaml_config.rb
mrbiggred: Good question and my answer is I don't know.  I'll do some testing to see what currently happens.  Then we can make a decision if the settings in `.standard.yml` should be overridden by the `.standard_todo.yml` settings.

mrbiggred: @searls what currently happens is the ignore list from the `.standard.yml` and `.standard_todo.yml` are combined into one list with the possibility of duplicates.  Duplicates are handled without any issue.

Actually the patterns in the ignore section for both files is translated into a Rubocop AllCops Exclude section, duplicates and all.  So really Rubocop has not problem with duplicate ignores and uses the broadest ignore pattern.

In the case where one has a broader ignore pattern then the other then the broader one wins.  For example the below will ignore all the files in `foo/bar`.  It does not matter which file the `foo/bar/**/*` is in.

```
# .standard.yml
ignore:
  - foo/bar/**/*
```

```
# .standard_todo.yml
ignore:
  - foo/bar/somefile.rb
```

I don't think we need to add more logic to detect duplicates or overlap with the files.  At least not right now.  If it becomes a problem we can add checks.

My vision for using Standard with the Todo file is as follows:

1) User has an existing project.
2) They install Standard, setup the initial `.standard.yml` config including ignoring folders such as `/tmp`.
3) They run Standard and see all the errors they can't fix right now.
4) They use Standard to generate the Todo file.  Now they can add Standard to their automated build with any new Standard errors caught by the automated build.
5) They remove files from the Todo file as they work on the application.  The warning message is continually printed to remind them about the Todo file.

What are your thoughts?

bhaibel:
Seagulling in to express how excited I am about this PR -- I'll feel a lot more able to use Standard in my existing projects once it lands.

searls:
@bhaibel thanks for the bump. I'll try to look at this tomorrow

lib/standard/loads_yaml_config.rb
searls: Thanks for this analysis!

searls:
Hey @mrbiggred, I just merged in master and can't get the build to pass, with this failure:

```
Standard::Runners::GenignoreTest#test_todo_generated [/home/circleci/repo/test/standard/runners/genignore_test.rb:25]:
--- expected
+++ actual
@@ -1 +1 @@
-{"ignore"=>["errors_one.rb", "errors_two.rb"]}
+{"ignore"=>[]}

```

Could you take a look at it?

mrbiggred:
@searls I'll investigate the error you are getting when trying to merge to master this morning and let you know what I find.

mrbiggred:
@searls I finally fixed the problem but I'm not sure the exact root cause.  The problem appears to be the Rubocop Config Store settings are leaking between tests.  If I run the failing tests all by it's self it works.  If the BuildConfigTests run before the GenignoreTests then the then GenignoreTests will fail.

It appears that Rubocop is somehow caching the Config Store settings.  Specifically the exclude settings to ignore the `tmp/**/*` files in the below:

```ruby
# test/standard/runners/genignore_test.rb

def config_store(config_root = nil, rubocop_yml = "config/base.yml", ruby_version = RUBY_VERSION)
  RuboCop::ConfigStore.new.tap do |config_store|
    config_store.options_config = path(rubocop_yml)
    options_config = config_store.instance_variable_get("@options_config")
    options_config["AllCops"]["TargetRubyVersion"] = ruby_version.to_f
    options_config["AllCops"]["Exclude"] |= standard_default_ignores(config_root)
  end.for("").to_h
end
```

Since the GenignoreTests puts it's test files in the `tmp` folder they would be ignored.  To fix the problem I changed the following:

```ruby
# test/standard/runners/genignore_test.rb

def create_config
  Standard::Config.new(nil, ["."], {}, Rubocop::ConfigStore.new)
end
```

to:

```ruby
# test/standard/runners/genignore_test.rb

def create_config(config_path)
  store = RuboCop::ConfigStore.new.tap do |config_store|
    config_store.options_config = config_path
  end

  Standard::Config.new(nil, ["."], {}, store)
end
```

That resets the Config Store so the the `tmp` folder is no longer excludes.  If you have any in-site into why this occurred and/or a better way to fix the test let me know.

searls:
@mrbiggred I'm happy to let the mystery be, as it were

searls:
Landed in 0.4.0! Thanks so much for your work on this and so that @bhaibel can use it!

I officially nominate you both to long-term maintenance of this feature because I don't suspect I'll ever use it 😄

mrbiggred:
@searls thanks for accepting my PR.  I'd be happy to fix any issues that arise so don't be afraid to ping me.  All the best.

PR #165, opened 2020-04-05
Document how to create the spring binstub

lakim:
Hey there,

Thanks for the great tool.
I wanted to speedup `standardrb`, especially when running in VS Code for linting and to format on save (needs https://github.com/rubyide/vscode-ruby/issues/548).

I created [spring-commands-standard](https://github.com/lakim/spring-commands-standard) to configure `spring` to generate the `bin/standardrb` binstub.

Example on a 13 lines file:

```shell
# Before
$ time bundle exec standardrb --fix path/to/file.rb
bundle exec standardrb --fix   1,33s user 0,35s system 102% cpu 1,650 total

# After
$ time bin/standardrb --fix path/to/file.rb
Running via Spring preloader in process 32323
bin/standardrb --fix   0,35s user 0,15s system 68% cpu 0,723 total
```

See the PR to update spring's README: https://github.com/rails/spring/pull/612

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,7 @@ the tool you're interested in. That aside, here are other known integrations asi
 from editor plugins:
 
 * [Pronto](https://github.com/julianrubisch/pronto-standardrb)
+* [Spring](https://github.com/lakim/spring-commands-standard)
 
 ## Contributing
 


searls:
Cool! 

Rubocop has some pretty sophisticated caching built-in (which standard benefits from for free), so that leads me to ask how does this performance improvement compare if you just run the same standard command twice (where the second invocation should be much faster) before calling it with Spring?

lakim:
Here's what I have:

```shell
$ spring stop
Spring stopped.

$ time bin/standardrb --fix path/to/file.rb
Running via Spring preloader in process 27904
bin/standardrb --fix   0,30s user 0,13s system 13% cpu 3,269 total

$ time bin/standardrb --fix path/to/file.rb
Running via Spring preloader in process 28086
bin/standardrb --fix   0,29s user 0,10s system 59% cpu 0,660 total
```

So the first binstub call is pretty expensive (around 2x more than a regular `bundle exec standardrb`.
But subsequent calls are roughly 2x faster than `bundle exec standardrb`.
No real benchmark here, but a few runs on my machine on a small file.

searls:
I can't tell from what you've posted if this benefit is due to rubocop caching or not, but I'm happy to link to your resource anyway if you think it's helping

lakim:
Great, thanks!

PR #168, opened 2020-04-20
Don't break do/end blocks with `rescue` nodes

artfuldodger:
For example, changing:

```ruby
lulz = [:a].map do
  :lol
rescue StandardError
  :nolol
end
```

to

```ruby
lulz = [:a].map {
  :lol
rescue StandardError
  :nolol
}
```

will result in a SyntaxError.

When running Standard against our codebase, this was the only
error that was introduced.

### DIFF
--- a/lib/standard/cop/semantic_blocks.rb
+++ b/lib/standard/cop/semantic_blocks.rb
@@ -114,6 +114,14 @@ module RuboCop::Cop
       end
 
       def correction_would_break_code?(node)
+        rescue_child_block?(node) || non_parenthesized_keyword_args?(node)
+      end
+
+      def rescue_child_block?(node)
+        node.children.any?(&:rescue_type?)
+      end
+
+      def non_parenthesized_keyword_args?(node)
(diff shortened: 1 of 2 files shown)


searls:
I think this is caused by the SemanticBlocks cop that we forked & own! [Code](https://github.com/testdouble/standard/blob/master/lib/standard/cop/semantic_blocks.rb) and [Test](https://github.com/testdouble/standard/blob/master/test/standard/cop/semantic_blocks_test.rb) here if you want to take a stab at starting with a failing test and fixing this!

artfuldodger:
@searls Hey thanks for the pointers, but did you see the code changes? I'm hoping I changed things in the right places..

searls:
LOL I am so sorry -- I thought this was an issue and not a PR. You're several steps ahead of me!

searls:
Released in 0.2.4! 🎊

artfuldodger:
Awesome, thank you!

PR #173, opened 2020-05-11
adding standardrb plugin for guard in the Readme

JodyVanden:
This PR adds a link to the repo of a gem using standardrb with the gem guard

Once installed, you can run guard in your terminal and it will output any errors on the files modified by the user.


searls:
Awesome, thanks!

JodyVanden:
woaw that was fast 👍 
thanks

PR #177, opened 2020-05-20
Print TODO warning as frontmatter

bhaibel:
When we print a long TODO file as the next-to-last message for both failures and passes, failure output is harder to interpret. The fact that both failures and passes "look the same" at the bottom of one's terminal is proving a source of confusion for some folks on my team.


searls:
Hey Betsy, I'll take your word for it, since I haven't seen the feature in action.

searls:
Landed in 0.4.3

PR #179, opened 2020-05-27
Add Envoy to the "Who uses Ruby Standard Style?" section

thomascchen:
Adds Envoy to the "Who uses Ruby Standard Style?" section.


searls:
🎉

PR #182, opened 2020-06-12
Add myRent to Standard users

dwightwatson:
We've just adopted Standard after spending way too much time coming up with our own configuration. So much better not having to worry about it any more!


searls:
Yay! Welcome 🙌

PR #196, opened 2020-09-01
Updates Style/Alias to prefer alias_method

### LINKED ISSUE #185: Proposal: change Style/Alias to prefer_alias_method (or disable)
Hi @searls and fellow standardrb-ers,

I was porting a gem to standardrb today when I noticed it auto-corrected an `alias_method` to `alias`.

I note that the following rules are currently in place for `Style/Alias`:

```yml
Style/Alias:
  Enabled: true
  EnforcedStyle: prefer_alias
```

Looking at the git blame for the `base.yml` file, it seems like this rule was never specifically tweaked, it was just brought in with the wholesale import of the original RuboCop defaults.

I would like to propose that we (a) change the enforced style to `prefer_alias_method`, or (b) disable the rule entirely.

As I was about to write my reasoning for this proposal, I actually found this recently opened issue on the Ruby Style Guide that sums up my thinking very well, so I'll point you to this first: https://github.com/rubocop-hq/ruby-style-guide/issues/821

Happy to discuss further, but I thought it'd be simplest just to point to that issue and gauge your feelings first.

Cheers!

jmkoni:
- Fixes #185
- Added line autocorrectable files to validate that the change happens

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -726,7 +726,7 @@ Standard/SemanticBlocks:
 
 Style/Alias:
   Enabled: true
-  EnforcedStyle: prefer_alias
+  EnforcedStyle: prefer_alias_method
 
 Style/AndOr:
   Enabled: true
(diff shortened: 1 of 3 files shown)


timriley:
Thanks @jmkoni for doing the follow-through work here! 🙇

PR #200, opened 2020-09-16
update rubocop and enable four new cops, re-enabling two that have been fixed

jmkoni:
Update Rubocop from [0.90](https://github.com/rubocop-hq/rubocop/releases/tag/v0.90.0) to [0.91](https://github.com/rubocop-hq/rubocop/releases/tag/v0.91.0), enabling:
  * [`Lint/UselessTimes`](https://github.com/rubocop-hq/rubocop/pull/8702)
  * [`Layout/BeginEndAlignment`](https://github.com/rubocop-hq/rubocop/pull/8628)
  * [`Lint/ConstantDefinitionInBlock`](https://github.com/rubocop-hq/rubocop/pull/8707)
  * [`Lint/IdentityComparison`](https://github.com/rubocop-hq/rubocop/pull/8699/)

re-enabling after bug fixes:
  * [`Bundler/DuplicatedGem`](https://github.com/rubocop-hq/rubocop/pull/8666)
  * [`Naming/BinaryOperatorParameterName`](https://github.com/rubocop-hq/rubocop/issues/8664)


searls:
This all seems totally legit!

Is there a matchin rubocop-performance release to bump?

jmkoni:
@searls not yet!

searls:
Awesome, I'm :+1: to merge & cut a release

will:
Hi, `Lint/ConstantDefinitionInBlock` breaks for [sorbet's enums](https://sorbet.org/docs/tenum) which look like this:

``` ruby
# (1) New enumerations are defined by creating a subclass of T::Enum
class Suit < T::Enum
  # (2) Enum values are declared within an `enums do` block
  enums do
    Spades = new
    Hearts = new
    Clubs = new
    Diamonds = new
  end
end
```

searls:
@will lol, Sorbet isn't making this project any easier, Will

will:
Yeah…

I figured it'd be an interesting thing to try on my new project, and like it's… okay. But not great. It's for sure caught some things I wouldn't have otherwise and the language server thing and vim+ale is neat. Though there are defiantly costs, and when I go over and do a little bit of Crystal I'm like "oh here is a type system that actually works with duck typing and isn't just annoying all the time" and then I get sad at sorbet.

searls:
@will does it make sense for us to either (1) go out of our way to ensure standard doesn't screw with sorbet? I haven't been keeping up with the Ruby 3 type protocol stuff, or (2) invest in an additional flag or mode that allows standard to dynamically switch to a different rubocop config for Sorbet projects (the same way we do for Ruby 1.8-2.3 projects)?

will:
>  Ruby 3 type protocol stuff,

I'm probably not all the way up to date, so this might be wrong, but my understanding is that ruby 3 will have out-of-line type info files that you can write called "rbs". Sorbet has a combination of in-line type signatures and also optionally out-of-line "rbi" files. Right now "rbs" and "rbi" are incompatible, but I think sorbet will at least optionally be change to support the "rbs" files.

as for the options, I think if anything (2). I have no idea how popular sorbet actually is, and so far the standard vs. sorbet problems have been due to very sorbet specific things rather than things that I think would be problems with ruby 3 typing. 

But I'd also be okay with option (3), make Will maintain a few ignores in `.standard.yml`. Let me know if you want me to stop reporting sorbet things, it's cool :)

jmkoni:
I'm open to you continuing to report with the chance that we might not fix it 😬 but if you open up an issue and it gets engagement, that at least can let us know how many other standard users are running into the same problem.

PR #201, opened 2020-09-17
Added config.ru files to include rb file list

megatux:
_config.ru_ files are pretty common on Ruby (web) projects


searls:
Sorry, this is for a rule that only applies to Bundler Gemfiles, so I don't think `config.ru` would be appropriate.

PR #203, opened 2020-09-18
of course rubocop-performance just updates 4 days later

jmkoni:
Was able to enable `Performance\Sum` due to a bug fix. Woo.


jmkoni:
Test is failing because `Performance/Sum` is not autocorrecting *even though it should*. I can switch up our tests to have it be in the unfixable spec, but I really want to know why this isn't correcting.

searls:
Can you reproduce it without involving standard so you can file it against the upstream repo?

jmkoni:
Opened up a bug with rubocop https://github.com/rubocop-hq/rubocop-performance/issues/171
I still think it's worth adding this rule because it does catch the issues, it just won't autocorrect most of them 😬

PR #209, opened 2020-10-06
Remove `Performance/Sum`

### LINKED ISSUE #208: Performance/Sum should be removed or should not be Autocorrect: true because it is not type-safe
In some of our code read docker-compose style durations like `1h30m` and we convert these to a single-unit duration.  A value must be valid and present and the conversion method enforces this.

During this conversion we want to reject an invalid unit, like `5n` (instead of `5m`).  Our lookup table returns `nil` for an invalid unit and attempts a multiplication which raises TypeError which we covert to an invalid duration.

If we have no units at all we depend on `Array#reduce(&:+)` on an empty Array to return `nil`, then pass that `nil` to Kernel#Integer which will raise TypeError, indicating an invalid (missing) duration.

As of this writing [autocorrect is enabled](/testdouble/standard/blob/c392b209d20aa77855e6df9f1e8c5d7b6fc85ede/config/base.yml#L706) in standard for Performance/Sum which changes the types for this method in an unsafe manner.  Instead of returning `nil` like `[].reduce(&:+)` which raises TypeError we get `0` from '[].sum` which Kernel#Integer does not raise a TypeError for.

Fortunately we have adequate tests to detect the type error that the suggestion introduced.

A performance improvement that introduces a bug is neither of those two things.

jmkoni:
- Fixes #208
- The rule isn't [working as intentended](https://github.com/rubocop-hq/rubocop-performance/issues/171)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## unreleased
+
+* Disable `Performance/Sum` because #208 and the lack of actual auto-correcting is also causing more trouble
+
 ## 0.7
 
 * Update rubocop from 0.91.1 to
--- a/config/base.yml
+++ b/config/base.yml
@@ -701,10 +701,6 @@ Performance/StartWith:
 Performance/StringReplacement:
   Enabled: true
 
-Performance/Sum:
-  Enabled: true
-  AutoCorrect: true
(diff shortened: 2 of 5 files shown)


searls:
LGTM

Issue #75, opened 2019-02-05
Proposal to reconsider trailing dot position


JoshCheek:
The setting we want to consider
=======================

So we have [this](https://github.com/testdouble/standard/blob/14da557e62e62e1a0f6639fd748dfc8731d83707/config/base.yml#L59-L61) config setting.

```yaml
Layout/DotPosition:
  Enabled: true
  EnforcedStyle: trailing
```

I'd like to make some arguments against it, provide counterarguments to the reason it's the default, and then wrap up with a summary.

Arguments against it
===============

### It causes code churn in git

@samphippen mentioned in the [podcast](https://twitter.com/josh_cheek/status/1092553489807560706) that he chooses syntactic preferences with the goal of reducing git churn. Lets analyze this setting with that metric in mind:

We append a line and look at the diff. What we see is that a leading dot causes a 1-line diff and a trailing dot causes a 3-line diff because we had to modify the end of the previous line. Not a huge deal, but it does increase cognitive overhead. You just have to be more on top of your game because the diffs have a less favourable signal to noise ratio.

![diff](https://user-images.githubusercontent.com/77495/52260458-c049d300-28eb-11e9-8861-16190eec1954.png)

### It's cumbersome to edit

Lets swap the last 2 lines instead. We find we can't just move lines around, because each line is not self-contained, it's partially on the line before it. Since the last line isn't before anything, it deviates from the pattern and we have to futz around with its trailing dot. The last line perpetually needs our attention.

![sib-alternative-4](https://user-images.githubusercontent.com/77495/52259423-482dde00-28e8-11e9-820f-29bb1928b472.gif)

### It puts important info way off at EOL

How one line relates to the next is pretty important, but there's lots of reasons a line might continue, 
in this example, maybe it's a method call, or maybe it's `line1 &&\n  line2` or `line1 -\n  line2` and so on. Important info is hard to access making this difficult to read.

![screen shot 2019-02-05 at 1 59 35 am](https://user-images.githubusercontent.com/77495/52260054-71e80480-28ea-11e9-80fa-e05ee5c77169.png)

Addressing the incumbent reasoning
===========================

### Why is it set this way?

The readme gives the reasoning ([here](https://github.com/testdouble/standard/commit/27c7c6211aa176108d75acdac9c8be20ab8db6ce#diff-04c6e90faac2675aa89e2176d2eec7d8R42)):

> **Trailing dots on multi-line method chains** - chosen because it makes copying lines into a REPL easier

[...]

PR #219, opened 2020-11-04
Change defaults for TrainingComma hash/array cops

skatkov:
I think that defaults for Trailing Commas cops should be switched to `EnforcedStyleForMultiline: comma`. 

> It helps with consistency, preventing copy-paste errors, and maintaining attribution (minimizing impact on diffs).

Quoting from /r/javascript: https://www.reddit.com/r/javascript/comments/6vqsjh/whats_with_the_trailing_commas/

Here are examples from rubocop:
https://docs.rubocop.org/rubocop/1.0/cops_style.html#styletrailingcommainarrayliteral
https://docs.rubocop.org/rubocop/1.0/cops_style.html#enforcedstyleformultiline-comma-3

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -1118,14 +1118,14 @@ Style/TrailingCommaInArguments:
 
 Style/TrailingCommaInArrayLiteral:
   Enabled: true
-  EnforcedStyleForMultiline: no_comma
+  EnforcedStyleForMultiline: comma
 
 Style/TrailingCommaInBlockArgs:
   Enabled: true
 
 Style/TrailingCommaInHashLiteral:
   Enabled: true
-  EnforcedStyleForMultiline: no_comma
+  EnforcedStyleForMultiline: comma
 
 Style/TrailingMethodEndStatement:
   Enabled: true


searls:
Sorry, we're keeping this `no_comma`. We'd initially set Standard up this way for about a year, but it was easily the most-complained-about setting. It also caused some obscure bugs & inconsistencies in edge cases around things like trailing commas in keyword args and hash parameters.

sdhull:
Can you elaborate a bit on these "obscure bugs & inconsistencies in edge cases"? The rationale for enforcing trailing commas in multiline literals largely echos the [rationale put forth](https://github.com/testdouble/standard/issues/75) for enforcing leading dots in multiline method chains. I don't have time to put together a sweet presentation like the linked issue but since this was the original setting I'm guessing some maintainers see the usefulness.

I'll just leave you with a quote from that thread
> The git churn should be taken seriously here.
> @strzibny

sdhull:
> trailing commas in keyword args and hash parameters

@searls I think I finally understand the point here. So what about setting preference for trailing commas for hash literals but no trailing comma for method invocations? (This is what we had set up at my previous company, which felt like a good balance.)

ElMassimo:
Is there any advantage to __not__ allowing trailing commas?

There are certainly advantages to [having them](https://maximomussini.com/posts/trailing-commas/) in multi-line arrays and hashes.

PR #220, opened 2020-11-05
upgrade rubocop to 1.2.0

jmkoni:
I want to enable the collection compact cop, but it seems like it might be a bit buggy right now.


searls:
LGTM, but don't feel like you need a review if the build passes and you think it'll work!

PR #222, opened 2020-11-17
Sort config YAML entries by key

amatsuda:
I found that the cops in config/base.yml are random ordered (or "almost sorted"), but it's easier to read for humans if the cops are sorted.

Here's a very roughly written sorting script, and the result is PRed.

```ruby
yml = File.read 'config/base.yml'

yml =~ /(.*)(AllCops:.*)/m
header, cops = $1, $2

sorted = cops.split("\n\n").sort_by { |cop| cop.lines.first }.join("\n\n")

File.write 'config/base.yml', header + sorted
```

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -173,6 +173,10 @@ Layout/IndentationConsistency:
   Enabled: true
   EnforcedStyle: normal
 
+Layout/IndentationStyle:
+  Enabled: true
+  IndentationWidth: ~
+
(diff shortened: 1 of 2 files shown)


searls:
Thanks so much for this, Akira! @jmkoni pointed this out recently and we haven't had the chance to do this ourselves.

Thank you for converting standard's config from Almost Sorted to Alphabetically Sorted 🔡

PR #225, opened 2020-11-30
updates to rubocop 1.4.2 and rubocop-performance 1.9.1

jmkoni:
Lots of updates over the holidays! There are three new performance rules that we are adding:

- Performance/BlockGivenWithExplicitBlock | https://github.com/rubocop-hq/rubocop-performance/pull/173
- Performance/ConstantRegexp | https://github.com/rubocop-hq/rubocop-performance/pull/174
- Performance/ArraySemiInfiniteRangeSlice | https://github.com/rubocop-hq/rubocop-performance/pull/175

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -621,10 +621,16 @@ Naming/VariableName:
   Enabled: true
   EnforcedStyle: snake_case
 
+Performance/ArraySemiInfiniteRangeSlice:
+  Enabled: true
+
 Performance/BigDecimalWithNumericArgument:
   Enabled: true
 
 Performance/BindCall:
+  Enabled: true
+
+Performance/BlockGivenWithExplicitBlock:
   Enabled: true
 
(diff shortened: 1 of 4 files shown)


config/base.yml
```diff
@@ -621,12 +621,18 @@ Naming/VariableName:
   Enabled: true
   EnforcedStyle: snake_case
 
+Performance/ArraySemiInfiniteRangeSlice:
+  Enabled: true
```
koic, line 625: FYI, `Performance/ArraySemiInfiniteRangeSlice` cop will be removed in the future, so it is recommended to disable.
https://github.com/rubocop-hq/rubocop-performance/issues/198

This cop is disabled by default and it's best not to list it in config/base.yml in case it's removed.

jmkoni: Thanks for the info!

Issue #249, opened 2021-01-21
Enhance our CI build to run against multiple Rubies


searls:
Because Standard is designed to run under whatever Rubies are supported by the version of rubocop we depend on and _against_ source listing compatibility all the way back to 1.8.7, I really would like to see our [GitHub Actions CI build](https://github.com/testdouble/standard/runs/1681563940) run with a few representative versions of MRI (perhaps oldest supported, most typical, and latest release)

Anyone familiar with using GitHub actions for Ruby CI interested in showing us how to set this up correctly?

anoymouscoder:
At present, I have seen some ruby versions in the ci file https://github.com/testdouble/standard/blob/master/.github/workflows/test.yml#L12-L15.  Does this issue already resolved?

searls:
Ah yep. I assumed the other PRs had marked themselves as resolving this one, thanks

PR #250, opened 2021-01-21
Draft: Setup matrix build

clupprich:
tbd

#249


searls:
That was quick! Looks great

PR #251, opened 2021-01-22
👷 Updates GitHub test action to use a cache and mutli-os

codenamev:
This helps speed up test runs by vendoring gems in a cache.  I've also introduced `os` into the matrix so that tests could easily run on multiple OSes as well.

Related to #249


searls:
👏 very nice!

PR #254, opened 2021-02-02
update rubocop version; added ruby 3.0 support

one0fnine:


jmkoni:
Hi @one0fnine! So we actually generally only need the config files for older versions. In addition, we don't want to disable SlicingWithRange for 3.0... that's the version that added that feature. Disabling it for 3.0 would be disabling it across the board. I'd be fine merging this if you fix the tests and remove that config file.

jmkoni:
Also, we would need an updated to CHANGELOG.md under `## unreleased`

one0fnine:
> Hi @one0fnine! So we actually generally only need the config files for older versions. In addition, we don't want to disable SlicingWithRange for 3.0... that's the version that added that feature. Disabling it for 3.0 would be disabling it across the board. I'd be fine merging this if you fix the tests and remove that config file.

Hi jmkoni! No problem, will do!

searls:
Hey @one0fnine, could you explain what you're trying to solver for with this PR? Standard already runs fine on Ruby 3 and should detect and target it. We have our own process for triaging Rubocop upgrades (it's a pretty involved, discussion-laden workflow, because it requires us to decide on what to do about every new and changed rule), and that's currently something we're working on apart from this PR

jmkoni:
Closing this because this is getting resolved in #258. In general, it's best to look at our open and approved tickets if you are interested in contributing. We try to upgrade pretty swiftly and usually have conversations about what we want to bring in, so it's usually not as small of a change as you might think.

PR #256, opened 2021-02-12
Add support for --stderr

knu:
cf. https://github.com/rubocop-hq/rubocop/pull/9043

I tried to add a test for this, but I couldn't because the implementation in Rubocop alters `$stdout`/`$stderr` and that conflicts with `do_with_fake_io`.


searls:
I'll trust you that this works, @knu! Thanks

PR #263, opened 2021-03-03
Replace SemanticBlocks with custom BlockDelimiters

### LINKED ISSUE #66: Semantic blocks are not enforced in hash values
Hi! I applied `standardrb` to a project (love it ❤️) and I noticed this code didn't get corrected with the semantic block cop:

```ruby
{
  foo: [].map do
    nil
  end,
}
```

Should this case be supported? Thanks!

### LINKED ISSUE #94: SemanticBlocks collides with Style/RedundantBegin
Whenever the `Standard/SemanticBlocks` rule autofixes a block which also matches rubocop's built-in `Style/RedundantBegin` rule, the autofix results in invalid ruby:

```ruby
def do_something(array)
  value = array.map do |item|
    begin
      maybe_fail(item)
    rescue
      handle_error
    end
  end
end
```

running `bundle exec standardrb --fix` on the above results in

```ruby
def do_something(array)
  value = array.map { |item|
    
      maybe_fail(item)
    rescue
      handle_error
    
  }
end
```

Have created a reproduction repo [here](https://github.com/wadetandy/standardrb-semantic-blocks-repro).

### LINKED ISSUE #246: False positive Standard/SemanticBlocks in method with rescue
The following code is ok according to Standard.
```ruby
class Foo
  def foo
    [1, 2].map { |value|
      value.to_s
    }
  end
end
```

However, when adding a rescue clause to the method, Standard gives a `Standard/SemanticBlocks` warning, although the code hasn't actually changed semantics.
```ruby
class Foo
  def foo
    [1, 2].map { |value|
      value.to_s
    }
  rescue
    raise
  end
end
```

Chaining a method after the `map` removes the warning again, so my guess is it has something to do with the "offending" block being the last before the `rescue`.

### LINKED ISSUE #255: Is there a way to disable the Semantic Block rule?
I get that the entire point of this project is to avoid these kinds of discussions and bikeshedding and this would undermine the intent of the project.

I also understand and agree with the intent behind the semantic block rule.

*but*

It seems impossible to actually get this rule right because there is way too much ruby code in the wild and this rule would have to know every case where it could mess up.

I opened #248 as an example of a class of issues, I was advised to change this specific instance, but I have encountered other cases that this doesn't address (this rule makes usage of the Dry::Initializer gem even uglier ;-) )

#109 was closed out because the specific issues mentioned in it had apparently been addressed, but I feel like the issue was also about the underlying problem.

I think #94 and #246 are also related to this.

I seriously want to thank all of the maintainers for making such a great library. I really appreciate your hard work.

If I have to– I guess I can just accept that standard will make my code ugly in places that goes against even its own (and (Jim Weirich's) intent, but it feels like this bummer is bigger than "i disagree with the code style choice" and actually one around correctness.

I understand if this issue gets insta-closed, but I do think there should be an open issue (this one or #109) until there is an explicitly stated decision the problem, rather than closing out issues around specific instances.

Seriously, Thanks again for this software, your time and effort are appreciated!

searls:
This change will move Standard from attempting to enforce Weirich-style semantic blocks, because there will always be edge-cases we can't cover automatically, given how expressive Ruby is and the fact that certain cases are genuinely ambiguous, such that an automated tool cannot reliably determine the authorial intent of a block.

Instead, the new rule will only enforce that single-line blocks be delimited with `{` and `}`, but otherwise allow users to use whatever multiline format they wish.

Resolves #255
Resolves #246
Resolves #94
Resolves #66


jmkoni (approved):
I think this is definitely the right move to make.

dnsco:
Thanks for doing this! And thanks again for the awesome project! 💖

R167:
Thank you! always had to disable because of certain edge cases and appreciate the more minimal form of enforcement now (which I really agree with!). Awesome project!!!

PR #269, opened 2021-03-04
Remove line from README about making rule changes pre-1.0.0

R167:
Since StandardRB just hit 1.0.0, remove the line from the README about changing rules pre-v1.0.0. (Under the "I disagree with rule X" section). At the very least, this should be updated.

Line in question: "**[NOTE: until StandardRB hits 1.0.0, the answer is yes! It just requires [opening an issue](https://github.com/testdouble/standard/issues/new) and convincing [@searls](https://twitter.com/searls) (the BDFNow) to make the change.]**"

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,11 +231,6 @@ style.
 
 ## I disagree with rule X, can you change it?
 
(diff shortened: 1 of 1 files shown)


searls:
Wups, must have missed that one. Thanks!

PR #270, opened 2021-03-05
Use RuboCop::CLI::Command::ExecuteRunner

nicksieger:
Preserves RuboCop exit status while using RuboCop logic for printing summary, either on stdout or stderr with `--stdin ... --stderr` options.

I ran into this issue while running a command like `cat file.rb | standardrb --no-color --stderr --fix --stdin file.rb`. I want the reformatted code on stdout with the warnings and summary on stderr. I don't know if the use of `Rubocop::Runner` was intentional, but using the execute runner works for my case.

### DIFF
--- a/lib/standard/cli.rb
+++ b/lib/standard/cli.rb
@@ -3,9 +3,6 @@ require_relative "loads_runner"
 
 module Standard
   class Cli
-    SUCCESS_STATUS_CODE = 0
-    FAILURE_STATUS_CODE = 1
-
     def initialize(argv)
       @argv = argv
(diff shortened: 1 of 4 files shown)


searls:
Great PR, @nicksieger! This looks great. As for:

>  I don't know if the use of Rubocop::Runner was intentional, but using the execute runner works for my case.

I wish I knew, too! It's been so long I really can't recall, and it seems to work the same locally, elsewise. Thank you!

searls:
Landed in 1.0.2

PR #272, opened 2021-03-06
Fix exit code for version and help runners

adam12:
Cast the runner's return value to `Integer` so we're going to be calling `exit` with an Integer. The original implementation had a ternary with explicit constants for the two exit codes. I could probably do something similar if the simple cast is insufficient.

Fixes this:

```
$ standardrb --help
... bunch of help output
TypeError: no implicit conversion from nil to integer
  /workspace/exe/standardrb:7:in `exit'
  /workspace/exe/standardrb:7:in `<top (required)>'
  /usr/local/bundle/bin/standardrb:23:in `load'
  /usr/local/bundle/bin/standardrb:23:in `<top (required)>'
```

```
$ standardrb --version
1.0.2
TypeError: no implicit conversion from nil to integer
  /workspace/exe/standardrb:7:in `exit'
  /workspace/exe/standardrb:7:in `<top (required)>'
  /usr/local/bundle/bin/standardrb:23:in `load'
  /usr/local/bundle/bin/standardrb:23:in `<top (required)>'
```

### DIFF
--- a/lib/standard/cli.rb
+++ b/lib/standard/cli.rb
@@ -11,7 +11,7 @@ module Standard
 
     def run
       config = @builds_config.call(@argv)
-      @loads_runner.call(config.runner).call(config)
+      @loads_runner.call(config.runner).call(config).to_i
     end
   end
 end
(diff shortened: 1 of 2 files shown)


searls:
Whoops! That's what I get for blindly trusting @nicksieger!

(jk, Nick's great and I should have thought to try these commands)

searls:
Landed in 1.0.3!

nicksieger:
Hey, the tests passed! Sorry about the breakage!

PR #280, opened 2021-03-24
Updates GitHub test action to use native gem cache

codenamev:
[The setup-ruby action was updated to allow installing and caching gems](https://github.com/ruby/setup-ruby#caching-bundle-install-automatically).  Since I introduced the caching mechanism before, I figured I'd updated it.

While this is much cleaner now, there is a caveat: only the latest stable versions of Ruby seem to build clean.  I left some commits to document this (the alternative that worked was excluding a `Gemfile.lock`).  If you are ok with dropping support for older Ruby versions from the build, or ignoring the `Gemfile.lock` from source control, then I can cleanup the commits on this. Otherwise I can just close this for now and revisit another time 😂

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
-        ruby-version: [2.5, 2.6, 2.7, 3.0]
+        ruby-version: [2.5, 2.7, 3.0]
 
     runs-on: ${{ matrix.os }}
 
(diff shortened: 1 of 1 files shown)


searls:
Sorry, I missed this -- could you take another look at getting this to work on 2.5? We just dropped 2.4 but want to make sure 2.5 is running in CI

codenamev:
Hey @searls, I've cleaned up the commits from earlier.  For some reason 2.6 is now showing some failures, so I opted to drop it since you don't have it configured currently.  Hope this is acceptable 😄

searls:
Thanks for being so responsive!

PR #281, opened 2021-03-24
remove EnforcedStyle: always on Style/AndOr to prevent unsafe change

andrewdang17:
Running `standardrb --fix` should only be making safe changes and `Style/AndOr` with `EnforcedStyle: always` is not safe. A common example is in controllers when you want to return early after a render or redirect

```
  redirect_to some_path and return
```

with `EnforcedStyle: always` it will autocorrect it to
```
  redirect_to some_path && return
```
which results in different behavior. 

I propose keeping the [default EnforcedStyle](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/AndOr) since it will still enforce using `&&` over `and` when you would want it to.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -747,7 +747,6 @@ Style/Alias:
 
 Style/AndOr:
   Enabled: true
-  EnforcedStyle: always
 
 Style/ArgumentsForwarding:
   Enabled: true


searls:
This… seems good? I honestly can't remember why we chose this over conditionals; I wonder if the default might have changed in rubocop over the last two years and we missed it? cc/ @jmkoni

jmkoni:
Yeah, it looks like it was added as this Rubocop PR two years ago (https://github.com/testdouble/standard/commit/ff58f60a871c952b9e2971843b6c78de3c4cb705). I have no issues with this. I'll go ahead and merge.

PR #295, opened 2021-05-12
Add Podia to list of companies

sebastians:
As of today, we are now using Standard.rb. No more bike shedding!


searls:
Welcome! Hope you're doing well @sebastians!

PR #302, opened 2021-06-04
Add '--generate-todo' to CLI help output

jalessio:
I couldn't remember the name of this CLI argument so I ran `standardrb --help` and noticed `--generate-todo` wasn't there. Any reason to exclude it from the `--help` output?

The help text is taken directly from the README for consistency:
https://github.com/testdouble/standard/blame/4af3f6c2a3e5c80450897165d1244a42957ddc4b/README.md#L94


searls:
Nope! It was just added later by an outside contributor so we missed it. Thanks!

PR #321, opened 2021-09-03
Add thoughtbot to user list

MatheusRich:
We're using standardrb both with our clients and internal tools. <3


searls:
Yay! Thank you!

PR #324, opened 2021-09-13
Change case indentation to match if/else

### LINKED ISSUE #322: Auto correct for case indentation with assignment is weird
Let's say we have the following file foo.rb:

```ruby
foo12345 = if 1.next
             1
           elsif 2.next
             2
           elsif 3.next
             3
           end

bar12345 = case 0.next
           when 1.next
             1
           when 2.next
             2
           when 3.next
             3
           end

p foo12345
p bar12345
```

`standardrb --fix foo.rb` changes this file to:

```ruby
foo12345 = if 1.next
  1
elsif 2.next
  2
elsif 3.next
  3
end

bar12345 = case 0.next
           when 1.next
             1
           when 2.next
             2
           when 3.next
             3
end

p foo12345
p bar12345
```

I think 

```ruby
bar12345 = case 0.next
           when 1.next
             1
           when 2.next
             2
           when 3.next
             3
end
```

should be

```ruby
bar12345 = case 0.next
when 1.next
  1
when 2.next
  2
when 3.next
  3
end
```

Is this intentional?

## Additional info

This change can fix this alignment but Layout/CaseIndentation is disabled explicitly.
I'm not sure this patch can be accepted or not.

```diff
diff --git a/config/base.yml b/config/base.yml
index b9828fc..81e7d87 100644
--- a/config/base.yml
+++ b/config/base.yml
@@ -62,7 +62,8 @@ Layout/BlockEndNewline:
 Layout/CaseIndentation:
   # Disabled because IndentOneStep can't be configured for one-liner cases. See:
   # https://github.com/rubocop-hq/rubocop/issues/6447
-  Enabled: false
+  Enabled: true
+  EnforcedStyle: end
 
 Layout/ClosingHeredocIndentation:
   Enabled: true
```

searls:
The case/when rules we had in place was inconsistent with the indentation rule for multi-line flow control and multi-line data structure literals (for which all subsequent lines are aligned with the end of the expression as opposed to the beginning). I lost track of this over time but definitely consider the current case shown in #322 to be a bug given how far off the mark it is.

```ruby
# bad
def count_carbs(food)
  carbs = case food
          when :pancakes
            23
          when :mushrooms
            4
          end
  carbs + 1
end

# good
def count_carbs(food)
  carbs = case food
  when :pancakes
    23
  when :mushrooms
    4
  end
  carbs + 1
end
```

This makes the rule significantly more strict than enforcement had been before, so I'm opening a PR for folks' opinions before merging, and suggest a minor bump.

Note, however, that this means that this PR would effectively ban what I personally see as a really aesthetically pleasing indentation pattern, however:

```ruby
# this is allowed but not enforced before patch
case fruit
  when :apple then "🍎"
  when :banana then "🍌"
end

# after merge the above will be autocorrected to this
case fruit
when :apple then "🍎"
when :banana then "🍌"
end
```

Losing the ability to optionally one-indent the `when` clauses is a bummer but it's strictly much more consistent to just require case/when to act like if/else, especially given that there's no enforcement to avoid the example at the top as pointed out in #322

Fixes #322

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -60,9 +60,8 @@ Layout/BlockEndNewline:
   Enabled: true
 
 Layout/CaseIndentation:
-  # Disabled because IndentOneStep can't be configured for one-liner cases. See:
-  # https://github.com/rubocop-hq/rubocop/issues/6447
-  Enabled: false
+  Enabled: true
+  EnforcedStyle: end
 
 Layout/ClosingHeredocIndentation:
   Enabled: true
--- a/test/fixture/cli/autocorrectable-bad.rb
+++ b/test/fixture/cli/autocorrectable-bad.rb
@@ -105,3 +105,13 @@ def bad_function(test: true, a:, b:)
     b
   end
 end
+
+def count_carbs(food)
+  carbs = case food
+          when :pancakes
+            23
+          when :mushrooms
+            4
+          end
+  carbs + 1
+end
--- a/test/fixture/cli/autocorrectable-good.rb
+++ b/test/fixture/cli/autocorrectable-good.rb
@@ -95,3 +95,13 @@ def bad_function(a:, b:, test: true)
     b
   end
 end
+
+def count_carbs(food)
+  carbs = case food
+  when :pancakes
+    23
+  when :mushrooms
+    4
+  end
+  carbs + 1
+end


camilopayan:
Looks good to me! And yea, the minor version bump sounds right.

calebthompson:
What about this option rather than lining up with the first line?

```ruby
# bad
def count_carbs(food)
  carbs = case food
          when :pancakes
            23
          when :mushrooms
            4
          end
  carbs + 1
end

# good
def count_carbs(food)
  carbs = case food
    when :pancakes
      23
    when :mushrooms
      4
  end
  carbs + 1
end

# also good
def count_carbs(food)
  carbs = case food
    when :pancakes then 23
    when :mushrooms then 4
  end
  carbs + 1
end
```

which would match up with what you've mentioned liking in:

```ruby
case fruit
  when :apple then "🍎"
  when :banana then "🍌"
end
```

I find those much more aesthetically pleasing than either of your initial two options, personally. They also make it easier to see at a glance where this "block" of code starts and ends.

delucas (approved):

searls:
@calebthompson you just described my ideal -- I don't think that RuboCop can currently support this

danielmorrison:
I don't have strong feelings here. I'm not sure I have the aesthetic appreciation for the indenting that you do, however (especially when it is a non `then` situation). 

My inclination would be to drop indenting and mimic `if`/`end`. I don't see the `case`/`when` being a "block" any more than I do for `if`/`else`. 🤷

searls:
> I don't see the `case`/`when` being a "block" any more than I do for `if`/`else`. 🤷

I think the case for this is that an `if` or `elsif` clause are two branches of the same kind, whereas `case` is followed by one kind of thing (an expression for compraison) and each `when` is followed by a different kind of thing (a condition matching that expression)

mtsmfm:
> you just described my ideal -- I don't think that RuboCop can currently support this

https://docs.rubocop.org/rubocop/cops_layout.html#layoutcaseindentation
`IndentOneStep: true` is the one I think

searls:
@mtsmfm Unfortunately that option is a bit all-or-nothing. What I'd like is to be able to allow 0 or 1 indents for `when` statements, and think the one step indent makes the most sense for one-line when statements (see this https://github.com/rubocop/rubocop/issues/6447#issuecomment-436011468 ), but I don't think I'd be comfortable forcing the `OneStepIndent` on people in all cases.

I'm going back and forth on whether to fork the rule to allow for this flexibility or think up a better option for this

mtsmfm:
If the new rule can be fixed automatically and has consistency, both rule (OneStepIndent: true and false) is acceptable for me.
That's why I choose this gem.

I'm happy if this PR is merged and released quickly since it's inconsistent now.

------

> What I'd like is to be able to allow 0 or 1 indents

> I'm going back and forth on whether to fork the rule to allow for this flexibility or think up a better option for this

BTW, I'm wondering if it can be autofix-able.

```ruby
case fruit
  when :apple then "🍎"
when :banana then "🍌"
end
```

Is the above example doesn't have any offence?
It depends on single line or multi line as you commented?
https://github.com/rubocop/rubocop/issues/6447#issuecomment-436011468

searls:
@mtsmfm I have not forgotten about this; just trying to make the time to investigate what it would take to update the underlying cop or potentially (😬) create our own to cover the nuance we'd ideally prefer

will:
Found a fun thing with this, though I'm pretty sure it's all just rubocop, so I opened an issue there https://github.com/rubocop/rubocop/issues/10528

searls:
@will neat! Thanks for filing it

PR #331, opened 2021-10-08
enable Lint/RedundantCopDisableDirective

grosser:
extra rubocop disables that are not needed should be removed

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -505,6 +505,9 @@ Lint/RaiseException:
 Lint/RandOne:
   Enabled: true
 
+Lint/RedundantCopDisableDirective:
+  Enabled: true
+
 Lint/RedundantRequireStatement:
   Enabled: true
 


searls:
Ok. Closing, but I'd probably support an internal fork of that cop since it seems better than not to clear out unnecessary comments, but I don't feel strongly enough about the upside to consider deprecating the "standard" comment—in general, many users of standard don't see or realize that it's running rubocop under the hood and shouldn't need to.

iainbeeston:
Interestingly it seems that @grosser 's example works correctly when checking for rules, but the autocorrect works differently and will remove rubocop disables for rules it doesn't know about. This can be an issue because standard doesn't define rules for things like metrics, and it will happily remove any metric disable comments that it sees. I'm not saying that there needs to be a code fix for this, but it's something to be aware of.

searls:
@iainbeeston that does seem like a legitimate issue. Especially because it's feasible that someone would run standard as well as a separate `rubocop` task for just some specific set of cops.

I think we should consider reverting this— @camilopayan?

iainbeeston:
@searls That’s exactly my issue! (Standard removes disable comments meant for rubocop).

My workaround is to disable this cop in standard by adding this to .standard.yml

```

ignore:
  - '**/*':
    - 'Lint/RedundantCopDisableDirective'

```

grosser:
if it's reverted then we a similar issue where running with rubocop does not remove unneeded disables ? (when using rubocop and standard)
... so maybe we need a cop that understands both to cleanly sole this ?

searls:
I don't think removing unnecessary disable comments is a priority. They're a no-op and they're not hurting anything. Based on what it sounds like here, a cop that would erroneously fail if someone's running two separate rubocop executions against a codebase or (worse) delete the comments in one execution and then cause surprising failures in a second would cause a real problem for people, however.

As a result I think we should revert this.

grosser:
Yeah true, the other issue is less annoying :)

camilopayan:
Yea, I agree that right now it's doing too much. I wouldn't want standardrb to start giving opinions on different rubocop configs with different use cases.

Issue #273, opened 2021-03-07
RUBOCOP conflicts 


MuhammedYousryKamel:
I have some Rubocop rules in the rubocop.yml file and need the standard gem to follow these rules so what should I do?
I need to change the rules inside the config/base.yml file

searls:
I'm sorry @MuhammedYousryKamel, but this usage isn't supported. Standard does not allow you to customize the rules, in order to encourage consistency across teams and projects.

You might find [this post](https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard) helpful, particularly the part about `inherit_gem`

Issue #271, opened 2021-03-05
Rubocop upset about BlockDelimiters change to config/base.yml


mepatterson:
I had everything working great with Standard + Rubocop + Rails. In my `.rubocop.yml`, I have this:
```
inherit_gem:
  standard: config/base.yml
```

Now, after updating versions, rubocop is blowing up with:

```
...gems/standard-1.0.2/config/base.yml: Standard/BlockDelimiters has the wrong namespace - should be Style
```

I use overcommit, so this is suddenly preventing me from committing due to the pre-commit hook failure. I don't really know how to proceed. Any ideas?

searls:
Hmm… sounds like it may be a bug in RuboCop. I suspect that RuboCop moved Style/BlockDelimiters from another section (Layout, maybe?) and as a result has migration warnings in there such that if a configuration contains a cop named BlockDelimiters in _any other_ section, it'll blow up and tell you to rename it

searls:
Should be worked-around in 1.0.4

swrobel:
Unfortunately I'm getting `Error: unrecognized cop Standard/BlockSingleLineBraces found in vendor/bundle/ruby/2.7.0/gems/standard-1.0.4/config/base.yml`

searls:
@swrobel could you please explain how you're using Standard such that this error is being raised?

swrobel:
Same way as the OP

Issue #211, opened 2020-10-14
standardrb compatibility with rubocop?


skatkov:
I remember that standard could be used as a configuration base for rubocop before. But these days, while going through README and documentation there is basically zero mentions about this option (and only note about base file).

Here is a configuration I found from "back in the days":
```
require:
  # Performance cops are bundled with Standard
  - rubocop-performance
  # Standard's config uses this custom cop,
  # so it must be loaded
  - standard/cop/semantic_blocks
  
inherit_gem:
  standard: config/base.yml
```

I tried using a similar configuration in a new project and it lead to all sort of surprising issues.

What is the current stance on this? Does standard still support this use-case, or we should just copy base configuration to the project and not use standard gem in our bundle?

searls:
We've never tried this or support it, because the point of Standard is that it not be configurable. I know some people have reported doing this with varying degrees of success (and more power to them!), but we're not considering making this work part of the project's purview

Issue #135, opened 2019-09-10
How should standard be used next to rubocop extensions?


pablocrivella:
I was wondering how should standard "integrate" with rubocop extensions like:

- https://github.com/rubocop-hq/rubocop-minitest
- https://github.com/rubocop-hq/rubocop-rspec
- https://github.com/rubocop-hq/rubocop-rails

searls:
I haven't looked at minitest or rspec, but I recently made the call to punt on rubocop-rails (for now). Since standard is more concerned with the formatting and structure of ruby-ruby than any particular library, I'm fine with punting on library-specific gems like this.

However, I could be convinced that how Standard _ought_ to behave is to curate & choose rules to enable from each of these, but only the most important lints (prioritizing safety over style), and to only activate them only in the presence of their corresponding `rubocop-foo` gem, as opposed to making any of them a hard dep of standard's. So for instance, if a user has `rubocop-rspec` in their Gemfile, our base configuration might enable certain cops bundled in that gem and start linting them, but otherwise we'd just carry on without it.

moger777:
Is it possible to opt into lint rules that are not part of the standard default? I know that defeats the purpose of having a standard to some extent but it is handy for say, opting into security linters that have less to do with style and more to do with preventing incorrect code.

searls:
@moger777 no, I don't think so. I think for this case you'd need to run rubocop and standard separately, and I'm not sure that's a bad thing (first of all, rubocop is already installed; second, I believe they share a cache; third, this is how StandardJS+eslint relate to one another in similar cases)

thijsnado:
I did the following which allows standard settings to be used in rubocop. Won't allow you to use other standard tooling but would get standard linting rules in addition to other linting rules:

```yml
require:
  - standard
  - rubocop-rails
  - rubocop-rspec

inherit_gem:
  standard: config/base.yml

Rails:
  Enabled: true
RSpec:
  Enabled: true
```

wilg:
@thijsnado I'm noticing doing that that I don't get all of the Standard formatting – for example, it doesn't enforce double quotes if I set my `.rubocop.yml` to:

```yaml
require:
  - standard
inherit_gem:
  standard: config/base.yml
```

thijsnado:
@wilg what version of rubocop/standard are in your Gemfile.lock?

wilg:
@thijsnado nvm, deleted, something else was wrong, sorry!

bf4:
[...]

PR #333, opened 2021-10-08
Document inherit_gem workflow

grosser:
afaik

pros:
- lots of people are using it judging from the issues/blogs I found
- it is a simple stepping-stone to adapting "standard"
- it plays nice with existing IDEs and pre-commit hooks

cons:
- more maintenance effort (but I think there is no plan to break it either since it has many users)

fixes https://github.com/testdouble/standard/issues/273 https://github.com/testdouble/standard/issues/30 https://github.com/testdouble/standard/issues/135 https://github.com/testdouble/standard/issues/211 https://github.com/testdouble/standard/issues/271
/cc @skatkov

### DIFF
--- a/README.md
+++ b/README.md
@@ -190,6 +190,22 @@ community conventions higher than personal style. This might not make sense for
 place for newbies. Setting up clear, automated contributor expectations makes a
 project healthier.
 
+## Usage via rubocop
+
+If you only want to use the rules and not the cli (to keep current IDE/tooling/workflow support).
+It needs to repeat `AllCops` because of a [rubocop issue](https://github.com/rubocop/rubocop/issues/10175).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -190,6 +190,22 @@ community conventions higher than personal style. This might not make sense for
...
+
+If you only want to use the rules and not the cli (to keep current IDE/tooling/workflow support).
+It needs to repeat `AllCops` because of a [rubocop issue](https://github.com/rubocop/rubocop/issues/10175).
+Change your `.rubocop.yml` to:
+
+```yaml
+require: standard
+
+inherit_gem:
+  standard: config/ruby-2.7.yml
```
searls, line 203: shouldn't this be `config/base.yml`?

PR #335, opened 2021-10-08
move default excludes into base.yml

grosser:
this makes inherit_gem workflow simpler and also makes the config easier to see when spelunking base.yml


searls:
Relevant issue to ignore behavior: https://github.com/testdouble/standard/issues/354

grosser:
- I don't think this would impact standards behavior.
- It makes the excludes visible in base.yml vs being hidden in code
- It allows inheriting the `Exclude` list when using `inherit_gem` workflow (vs having to define it manually)

Issue #332, opened 2021-10-08
All cops should be enabled or disabled


grosser:
Atm lots of good rules are disabled by accident for example `Lint/RedundantCopDisableDirective`.
As new rules are added to rubocop, they should be explicitly enabled or disabled here.

so I'd propose:
- remove `DisabledByDefault: true`
- add `NewCops: enable`
- add a test that says "every rule in rubocop needs to be either enabled or disabled"
- each rubocop version upgrade decides if the new rule is enabled or disabled

related issue: https://github.com/testdouble/standard/issues/24

/cc @elia @searls

grosser:
made a PR to fix this https://github.com/testdouble/standard/pull/336

PR #336, opened 2021-10-10
configure all cops + make ci check that they are configured

grosser:
fixes https://github.com/testdouble/standard/issues/332


test/standard_test.rb
```diff
@@ -7,4 +7,32 @@ def test_loads_stuff
...
+    extra = (configured - allowed).grep(/\AA-Z/) # ignore "require"
+    return if missing == [] && extra == []
+
+    if missing.any? && !ENV["CI"]
+      # rewrite file to add missing sections
+      separator = "\n\n"
+      sections = File.read(base).strip.split(separator)
+      sections += missing.map { |key| "#{key}:\n  Enabled: false" }
+      sections = sections.first(1) + sections[1..-1].sort # keep require in the front
+      File.write(base, sections.join(separator)) + "\n"
```
al2o3cr, line 31:
```suggestion
      File.write(base, sections.join(separator) + "\n")
```
This is what `Lint/Void` was complaining about.

jasonkarns:
Just noticing something while researching some things: in the body of #332, @grosser's proposal suggested that `NewCops: enable` be added. However, it is not added in this PR. (And I don't see any churn of that setting in the commit history.)

Was that intended to be added? Any rationale for omitting it?

grosser:
yeah good find, I think I just forgot it (or at least don't remember why it is not there)
afaik should be added + new ignores for whatever is undesired

jasonkarns:
@grosser thanks for confirming my suspicion! :smiling 

With rubocop's current behavior for an un-specified `NewCops` generating warnings, I think that gives us the desired behavior. The warning being the prompt to ensure all new cops are configured explicitly.

Issue #334, opened 2021-10-08
What's the correct ruby_version key to use with standardrb and ruby-next?


todd-a-jacobs:
## Problem

I'm trying to use standardrb with the ruby-next gem, which basically polyfills or backports features from later versions of Ruby. I opened a [related issue with ruby-next](https://github.com/ruby-next/ruby-next/issues/82) because standardrb (and RuboCop itself) seems to need a TargetRubyVersion, but ruby-next sets `next` when:

``` yaml
# .standard.yml
require:
  - ruby-next/rubocop
```

I've tried setting `ruby_version: 'next'` in .standard.yml, as well as simply setting it to `3.0`, but I still get errors from standardrb such as:

> /Users/$USER/.rvm/rubies/ruby-2.7.3/lib/ruby/2.7.0/rubygems/version.rb:215:in `initialize': Malformed version number string next (ArgumentError)

or the following out of RuboCop:

```
(Using Ruby 3.01 parser; configure using TargetRubyVersion parameter, under AllCops`)
```

What would be a useful value for the ruby_version key in standardrb when using ruby-next?

PR #342, opened 2021-10-13
Support ruby-next versioning

tuxagon:
The goal of this change was to explore what would be involved in supporting `ruby-next` versioning from https://github.com/testdouble/standard/issues/334.

**Disclaimer**: I had never used `ruby-next` before, so there may need to be a more experienced confirmation beyond the basics I did here.

I'm not sure how much `ruby-next` is used, but my thoughts about this particular issue are more about compatibility with rubocop rather than supporting `ruby-next`. Since standard is a thin-wrapper to prevent bikeshedding with configuring rubocop, a change like this makes sense to me to ensure surprising crashes like this caused are handled better. Any discussion about this is welcome.

The essentials here is that `ruby-next` requires that the `TargetRubyVersion` setting is given a value of `next`, but `standard` used `Gem::Version.new` in a few places which requires a traditional version number. This PR changes that to allow an unconventional target version that isn't as rigid.

## Development Setup with Ruby Next

```yaml
# .rubocop.yml
require: standard

inherit_gem:
  standard: config/base.yml

AllCops:
  DisabledByDefault: true
```

```yaml
# .standard.yml
ruby_version: next
```

```ruby
# Gemfile
source "https://rubygems.org"

git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }

gem "ruby-next"
gem "standard", git: "https://github.com/testdouble/standard.git", branch: "334-support-require-for-ruby-next"
```

```ruby
# main.rb (from the ruby-next README)
require "ruby-next"

using RubyNext

def greet(val) =
  case val
    in hello: hello if hello =~ /human/i
      "🙂"
    in hello: "martian"
      "👽"
    end

puts greet(hello: "martian")
```

- [Ruby Next README](https://github.com/ruby-next/ruby-next)
- [Ruby Next Rubocop Setup](https://github.com/ruby-next/ruby-next#rubocop)

## Questions

- ~I didn't take a TDD approach to experimenting to this solution, so I'm curious as to where I should put the tests for verifying a non-standard version number.~

- It may not be necessary, but do you know why `- ruby-next/rubocop` cannot be added alongside `standard` in `.rubocop.yml` like
```yaml
require:
- standard
- ruby-next/rubocop
```

## Todo

- [ ] Add some documentation once everything is resolved

### DIFF
--- a/lib/standard/creates_config_store/assigns_rubocop_yaml.rb
+++ b/lib/standard/creates_config_store/assigns_rubocop_yaml.rb
@@ -10,7 +10,11 @@ class Standard::CreatesConfigStore
     private
 
     def rubocop_yaml_path(desired_version)
-      file_name = if desired_version < Gem::Version.new("1.9")
+      default = "base.yml"
+
+      file_name = if !Gem::Version.correct?(desired_version)
+        default
+      elsif desired_version < Gem::Version.new("1.9")
         "ruby-1.8.yml"
       elsif desired_version < Gem::Version.new("2.0")
         "ruby-1.9.yml"
@@ -25,7 +29,7 @@ class Standard::CreatesConfigStore
       elsif desired_version < Gem::Version.new("3.0")
         "ruby-2.7.yml"
       else
-        "base.yml"
+        default
       end
 
       Pathname.new(__dir__).join("../../../config/#{file_name}")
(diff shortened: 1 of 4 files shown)


tuxagon:
The tests I have in mind include explicitly ensuring version is set correctly for both conventional and unconventional versions.

tuxagon:
Added some tests to verify standard can set both types of ruby versions for rubocop.

Issue #330, opened 2021-10-06
standardrb reports a useless warning when .standard_todo.yml is empty


todd-a-jacobs:
Standard reports a warning about ignoring files even when the todo file has no ignores. For example:

``` yaml
# Auto generated files with errors to ignore.
# Remove from this list as you refactor files.
---
ignore: []
```

Running standard results in a loud warning, but lists no files because nothing is being ignored. It just complains and then returns you to the prompt. For example:

``` bash
$ standardrb
WARNING: this project is being migrated to standard gradually via `/Users/$USER/foo/.standard_todo.yml` and is ignoring these files:
$ 
```

The expected behavior would be for standardrb to avoid warnings when the todo file is present but empty. Removing the ignore line from the YAML makes no difference, so the behavior seems to be triggered simply by the presence of the todo file. Since this file is likely to be checked into source control and therefore may be empty but present during development, the warning should be skipped when appropriate.

PR #343, opened 2021-10-13
Prevent todo warning for empty file list

tuxagon:
Per https://github.com/testdouble/standard/issues/330

The `Standard::Formatter` better handles the situation where a todo file exists, but no files are actually listed within.

## Development Setup

Run `standardrb --generate-todo` to generate a `.standard_todo.yml` file. On `main`, you'll see the warning without files listed while on this branch, the warning does not show.

### DIFF
--- a/lib/standard/formatter.rb
+++ b/lib/standard/formatter.rb
@@ -87,7 +87,7 @@ module Standard
       return if todo_ignore_files&.any?
 
       output.print <<-HEADER.gsub(/^ {8}/, "")
-        Congrats, you're done! Delete `#{todo_file}` in celebration.
+        Congratulations, you've successfully migrated this project to Standard! Delete `#{todo_file}` in celebration.
       HEADER
     end
 
--- a/test/standard/formatter_test.rb
+++ b/test/standard/formatter_test.rb
@@ -30,7 +30,7 @@ class Standard::FormatterTest < UnitTest
     @subject.finished([@some_path])
 
     assert_equal <<-MESSAGE.gsub(/^ {6}/, ""), @io.string
-      Congrats, you're done! Delete `.standard_todo.yml` in celebration.
+      Congratulations, you've successfully migrated this project to Standard! Delete `.standard_todo.yml` in celebration.
     MESSAGE
   end
 


searls:
Shouldn't this warn that an empty todo list is present and ought to be deleted? It doesn't make sense for an empty todo list to live in version control forever.

tuxagon:
Hmm 🤔. As I was diving around the code tonight, getting a feel for it, I didn't really question _why_ you'd want to keep a `.standard_todo.yml` file around, so that's a good question. Instead, I agree, the change's goal _should_ be around improving the message provided when this situation arises.

Shouldn't take long to rework and get this updated tomorrow sometime.

lib/standard/formatter.rb
```diff
@@ -66,12 +66,18 @@ def print_todo_warning
...
 
-      output.print <<-HEADER.gsub(/^ {8}/, "")
-        WARNING: this project is being migrated to standard gradually via `#{todo_file}` and is ignoring these files:
-      HEADER
-
-      todo_ignore_files.each do |f|
-        output.printf("  %s\n", f)
+      if todo_ignore_files.size.zero?
+        output.print <<-HEADER.gsub(/^ {10}/, "")
+          WARNING: this project appears to already be migrated to standard! If that is true, `#{todo_file}` can be removed. Otherwise, files will need added to the `ignore` list inside `#{todo_file}`.
```
searls, line 71: > files will need added to the `ignore`

This phrase confuses me a bit -- if zero violations occur shouldn't you be removing files from the ignore list (to generate more todos), if any?

tuxagon: From my understanding, if someone wants to ignore certain files, then they need to add those files to the list. The idea behind the second part of the message is to capture directing the user on what to do next after generating the todo file.

searls: Ah, I see -- but what message should/will print if `standard` exits cleanly and this file also still exists? Shouldn't it just be "delete the file"?

tuxagon: By clean exit, do you mean the situation where rubocop finds no issues? In that case, I think we could safely delete the file for them. The one caveat I'm considering, however, is to potentially keep the file and print the warning _unless_ `--fix` is applied. If `--fix` is provided then standard could remove the file.

With that said, `--fix` is intentionally changing project code whereas `.standard_todo.yml` isn't exactly "project-specific" in the same way. It's more a metadata file for standard. I can see an argument both ways.

What are your thoughts?

lib/standard/formatter.rb
```diff
@@ -75,6 +79,18 @@ def print_todo_warning
...
 
+    def print_todo_congratulations
+      todo_file = options[:todo_file]
+      return unless todo_file
+
+      todo_ignore_files = options[:todo_ignore_files]
+      return if todo_ignore_files&.any?
+
+      output.print <<-HEADER.gsub(/^ {8}/, "")
+        Congrats, you're done! Delete `#{todo_file}` in celebration.
```
searls, line 90: Maybe to be clearer as to what just happened:

> Congratulations, you've successfully migrated this project to Standard!…

PR #345, opened 2021-10-14
enable Style/NumericPredicate with comparison

grosser:
https://docs.rubocop.org/rubocop/1.0/cops_style.html#stylenumericpredicate

Prefer `foo == 0` over `foo.zero?` since anyone can understand the first one and it avoids bugs with foo being `nil` (same for .positive? and .negative?)

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -1470,7 +1470,8 @@ Style/NumericLiterals:
   Enabled: false
 
 Style/NumericPredicate:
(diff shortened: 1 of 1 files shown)


searls:
I personally am very skeptical of any cops that enforce using one API over another like this without a really compelling safety or performance rationale (beyond the potential it could be nil)

PR #348, opened 2021-10-15
actionable code coverage

grosser:
- almost no runtime overhead
 - tells you when code is not covered (instead of generating a report nobody looks at)
 - encourages that test location/name matches file under test
 - supports branch coverage `a ? b : c` == 2 branches

before:
```
Coverage report generated for Unit Tests to coverage. 234 / 324 LOC (72.22%) covered.
```

after:
```
lib/standard/rubocop/ext.rb new uncovered lines introduced (1 current vs 0 configured)
Lines missing coverage:
lib/standard/rubocop/ext.rb:5
```

I can do the remaining files, but wanted to get a 👍 / 👎 before I put in more work.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -7,7 +7,4 @@ gem "minitest", "~> 5.0"
 gem "pry"
 gem "rake", "~> 13.0"
 gem "gimme"
-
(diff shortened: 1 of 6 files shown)


searls:
I'm sorry I don't think I'd like to add this right now.

PR #356, opened 2021-12-01
add ruby 3.0

grosser:


searls:
Could you help by providing some context to what this PR is  trying to do? IIRC, Style/HashExcept is already enabled in config/base.yml, which all projects targeting Rubies 2.5 & up will be loaded by standard

grosser (commented):
I need standard to work on 3.0 ... soI copied 2.7 🤷

grosser (commented):
updated to remove it from 2.7 and 3.0

searls:
I don't understand. As it is, Standard works on Ruby 3.

grosser:
it does not have a file for 3.0 so I'd have to use `base.yml` which might break when standard is updated, so I was thinking referencing 3.0 would be a nice consistent way of handling things (with all the 2.x versions flying around)

grosser:
making users use `standard: config/3.0.yml` is a nice reminder for them to upgrade when ruby version changes and gives us a chance here to have different configs per minor versions

camilopayan:
For all intents and purposes, the `base.yml` file acts as the "latest" version for rubocop config, so we add version config files when there's a configuration or cop that only works on the latest. Adding a 3.0 file doesn't really put it in front of the user in a meaningful way (since they just run standard) so I'm gonna go ahead and close this. Thanks, though!

PR #357, opened 2021-12-01
Update rubocop-performance gemspec

Kuraiou:
# What

* use 1.12.x for rubocop-performance.

# Why

It's the most recent version and bugfix versions, generally speaking, should be allowable (i.e. use `~>` instead of `=`).

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -923,6 +923,9 @@ Performance/CollectionLiteralInLoop:
 Performance/CompareWithBlock:
   Enabled: true
 
+Performance/ConcurrentMonotonicTime:
+  Enabled: true
+
 Performance/ConstantRegexp:
   Enabled: true
 
--- a/lib/standard/version.rb
+++ b/lib/standard/version.rb
@@ -1,3 +1,3 @@
 module Standard
-  VERSION = Gem::Version.new("1.4.0")
+  VERSION = Gem::Version.new("1.4.1")
 end
--- a/standard.gemspec
+++ b/standard.gemspec
@@ -20,5 +20,5 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "rubocop", "1.22.3"
-  spec.add_dependency "rubocop-performance", "1.11.5"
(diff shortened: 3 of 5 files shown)


standard.gemspec
```diff
@@ -20,5 +20,5 @@ Gem::Specification.new do |spec|
   spec.require_paths = ["lib"]
 
   spec.add_dependency "rubocop", "1.22.3"
-  spec.add_dependency "rubocop-performance", "1.11.5"
+  spec.add_dependency "rubocop-performance", "~> 1.12"
 end
```
searls, line 24: We have made an intentional decision to lock to specific releases of rubocop and rubocop-performance, in order that standard's behavior at any given version is reproducible in all cases for a given target Ruby. If you'd like to bump to the latest rubocop-performance release, that's great (we just hadn't gotten to it yet), but please lock to the specific version.

config/base.yml
```diff
@@ -923,6 +923,9 @@ Performance/CollectionLiteralInLoop:
 Performance/CompareWithBlock:
   Enabled: true
 
+Performance/ConcurrentMonotonicTime:
+  Enabled: true
```
searls, line 927: This cop does not support Rubies older than 2.3 (based on [this issue linked from the rubocop PR](https://github.com/rails/rails/pull/43502)), so to enable it in the base means it should be disabled in the 2.2 config file.

lib/standard/version.rb
```diff
@@ -1,3 +1,3 @@
 module Standard
-  VERSION = Gem::Version.new("1.4.0")
+  VERSION = Gem::Version.new("1.4.1")
```
searls, line 2: Generally, it's considered bad form to bump the version number of a library in a PR, as it's the maintainer's responsibility to manage what goes into each release, in addition to version control, packaging, and publishing of each release (which, if you see the git log of the `version.rb` file, you'll see we're very particular about). As a result, it can sometimes come across as pushy or forward to bump the version of a library in a PR (the implication being an imperative "cut a release as soon as this is merged just for this change")

searls (changes requested):

camilopayan:
Should be covered in the latest v1.5.0 version.

PR #362, opened 2021-12-03
Add Babylist to "Who uses Ruby Standard Style?" list

jalessio:
- Alphabetizes the "Who uses Ruby Standard Style?" list, with the exception of leaving Test Double at the top
- Adds Babylist to the list

Let me know if you'd prefer I just add Babylist to the bottom of the list instead of sorting alphabetically.


searls:
it's all good! alphabetizing makes sense at this point ❤️

PR #364, opened 2021-12-07
Update section about usage via rubocop

joergschiller:
`DisabledByDefault: true` was removed from config/base.yml in https://github.com/testdouble/standard/commit/2445c16420428d75bd68208f069fc02d069c5a63. According to https://github.com/testdouble/standard/issues/332 this was intentionally.

I think it's no longer necessary to add it to the rubocop config file here.


searls:
Thanks!

PR #379, opened 2022-01-14
Ensure that CI runs a 3.0.x version

petergoldstein:
Because YAML truncates trailing zeroes in numerical values, an unquoted 3.0 is treated a 3.  This loss of precision means that the actual version of Ruby loaded will be the most recently release 3.x version - at this time 3.1.0.

Quoting the '3.0' restores the precision.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       matrix:
         os: [ ubuntu-latest ]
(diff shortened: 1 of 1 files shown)


petergoldstein:
You can see this by looking at the details for a recent CI action:

Screen Shot 2022-01-13 at 9 56 40 PM

Note that the '3.0' in the left nav is replaced by a '3' and the the loaded Ruby is 3.1.0

searls:
TIL!

Issue #385, opened 2022-02-03
Style/RedundantBegin breaks Ruby 2.4 and below


ivoanjo:
Howdy, and thanks for standard! I really like the tool and I added it [on one of Datadog's Ruby gems](https://github.com/DataDog/libddprof/tree/main/ruby).

Today I ran into an issue with one of our RSpec tests. I wrote down

```ruby
    after do
      begin
        FileUtils.remove_dir(temporary_directory)
      rescue Errno::ENOENT => _e
        # Do nothing, it's ok
      end
    end
```

which got autofixed to 

```ruby
    after do
      FileUtils.remove_dir(temporary_directory)
    rescue Errno::ENOENT => _e
      # Do nothing, it's ok
    end
```

...which does not work on Ruby 2.4:

```ruby
Failure/Error: __send__(method, file)

SyntaxError:
  /Users/ivo.anjo/datadog/libddprof/ruby/spec/libddprof_spec.rb:27: syntax error, unexpected keyword_rescue, expecting keyword_end
      rescue Errno::ENOENT => _e
            ^
  /Users/ivo.anjo/datadog/libddprof/ruby/spec/libddprof_spec.rb:27: syntax error, unexpected =>
      rescue Errno::ENOENT => _e
                             ^
  /Users/ivo.anjo/datadog/libddprof/ruby/spec/libddprof_spec.rb:97: syntax error, unexpected keyword_end, expecting end-of-input
```

I have my `.standard.yml` set to `ruby_version: 2.1`. The cop being triggered is ``Style/RedundantBegin: Redundant `begin` block detected.``.

PR #390, opened 2022-02-14
Removes RedundantBegin from Ruby 2.4 and below

camilopayan:
Rubocop has dropped support for Ruby <2.5 so this cop no longer
functions safely on Ruby 2.4 and below.


camilopayan:
This addresses #385

ivoanjo:
Thank you for fixing this! :)

PR #402, opened 2022-03-21
Use bundle add instead

glaucocustodio:
As per https://github.com/rubygems/rubygems/pull/5337, we can simplify the steps of adding a gem.


searls:
Thanks for the suggestion, but adding the Gemfile stanza is pretty standard for READMEs and I have never seen people use `bundle add` (nor would I use it, since I'm pretty precious about how I format my Gemfiles). It would preclude folks using alternate tools like [gel](https://github.com/gel-rb/gel)

PR #408, opened 2022-04-23
enforce :"foo-bar" over :'foo-bar' to be consistent with other doubl…

grosser:
…e-quoting rules

```
-'Impersonate-User': IMPERSONATE_USER,
+"Impersonate-User": IMPERSONATE_USER,
```


grosser:
updated

Issue #133, opened 2019-09-03
Trailing comma rule results in `},])` contortion


searls:
Standard keeps slamming the final `,` into this and I hate it:

```ruby
    Item.find_by_sql([sql, {
      user_id: user.id,
      answer: answer,
    },]).uniq
```

There's no way anyone thinks `},])` is preferable in this case, right?

pablocrivella:
Yup, I don't like it either 😬 
Would the solution be only adding the comma to the last element of a hash?

searls:
Yes, one would think

GeoffAbtion:
`},]) ` is undesirable but I think it is a symptom of another problem, you're trying to wrap a hash in a strange way. Of course that is going to look wrong.

These all look reasonable:

1)
```ruby
Item.find_by_sql(
  [
    sql,
    {
      user_id: user.id,
      answer: answer,
    },
  ]
).uniq
```
2)
```ruby
Item.find_by_sql([
  sql,
  {
    user_id: user.id,
    answer: answer,
  },
]).uniq
```
3)
```ruby
Item.find_by_sql([
  sql,
  {user_id: user.id, answer: answer},
]).uniq
```
4)
```ruby
Item.find_by_sql([sql, {user_id: user.id, answer: answer}]).uniq
```

2) is probably what I'd want to see.

jmkoni:
Pretty sure we fixed this in https://github.com/testdouble/standard/commit/a7156b14528f4b171567c528f6a2434a43b6e54a

Issue #159, opened 2020-02-23
Reconsider position on adopting trailing commas


penelopezone:
Hi Folks,

I am unsure trailing commas are the right solution, and I wrote a detailed report on the [`rubyfmt`](https://github.com/penelopezone/rubyfmt/issues/154) issue tracker that I'd love some discussion on.

searls:
So @penelopezone has decided in https://github.com/penelopezone/rubyfmt/issues/154 to eliminate trailing commas from rubyfmt.

It's long been a goal of both of us to keep these two projects in relative harmony with one another (ideally, running the same source listing through both tools would produce consistent results, such that standard could be winnowed down to just linting tasks and rubyfmt could take over for style/layout on a file-save-hook), and as a result I'm inclined to cascade her decision to standard as well.

Other reasons I support this decision:

* Selfishly, I don't personally like trailing commas so of course I don't mind it (see my comment in the above linked thread)

* Rubocop's defaults don't append trailing commas, and we've seen a handful of gnarly issues caused by our particular configuration including them

searls:
Fixed by a7156b14528f4b171567c528f6a2434a43b6e54a and released in v0.2.0

thorncp:
Looks like the current config doesn't take a stance on method calls. E.g. both of these are considered valid:

```ruby
doit(
  foo: "bar",
  baz: "qux"
)

doit(
  foo: "bar",
  baz: "qux",
)
```

Is that intentional?

searls:
Can you help me find the rubocop rule that would disallow the latter?

thorncp:
How about a PR? https://github.com/testdouble/standard/pull/160

PR #453, opened 2022-08-31
Disable TrailingComma* cops

casperisfine:
These cops used to be configured with `EnforcedStyleForMultiline: consistent_comma` but it was changed in https://github.com/testdouble/standard/commit/a7156b14528f4b171567c528f6a2434a43b6e54a because it can cause some weird contortions, see: https://github.com/testdouble/standard/issues/133

I totally agree that this cop resulting in such horrible contorsion is broken, but I don't understand why that would justify to switch to a different style rather than to just disable the cop.

Trailing comma in multiline hashes & co have the major advantage of reducing diff noise when adding or removing an element.

NB: I'm not too sure if it's ok to open PRs to argue about style, but since this used to be the preferable style and was changed for a technicality, I hope it's ok.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -1816,19 +1816,16 @@ Style/TrailingBodyOnModule:
   Enabled: true
 
 Style/TrailingCommaInArguments:
-  Enabled: true
-  EnforcedStyleForMultiline: no_comma
+  Enabled: false
 
 Style/TrailingCommaInArrayLiteral:
-  Enabled: true
-  EnforcedStyleForMultiline: no_comma
+  Enabled: false
 
 Style/TrailingCommaInBlockArgs:
   Enabled: true
 
 Style/TrailingCommaInHashLiteral:
-  Enabled: true
-  EnforcedStyleForMultiline: no_comma
+  Enabled: false
 
 Style/TrailingMethodEndStatement:
   Enabled: true
--- a/lib/standard/version.rb
+++ b/lib/standard/version.rb
@@ -1,3 +1,3 @@
 module Standard
-  VERSION = Gem::Version.new('1.16.1')
+  VERSION = Gem::Version.new("1.16.1")
 end
--- a/test/fixture/cli/autocorrectable-good.rb
+++ b/test/fixture/cli/autocorrectable-good.rb
@@ -6,12 +6,12 @@ STUFF = [
   1,
   3,
   4,
-  5
+  5,
 ]
 
 THINGS = {
   oh: :io,
-  hi: "neat"
+  hi: "neat",
 }
 
 class Something
@@ -70,7 +70,7 @@ class AlignyStuff
     thing_1: 0,
     thing_2: 1,
     longer_thing: 2,
-    even_longer_thing: 3
+    even_longer_thing: 3,
   }
 
   def setup_fog_credentials(config)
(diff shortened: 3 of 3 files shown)


casperisfine:
Also: I'm not very good with the rubocop codebase, but I wonder if it would be worth improving the upstream Cop so that it doesn't generate this kind of contraptions (I may try my hand at it).

searls:
Hi @casperisfine, thank you for the contribution.

The context on why we chose to ditch trailing commas is broader than the issue linked above. Here's [the issue](https://github.com/testdouble/standard/issues/159) that we tracked the changed to and the discussion in [rubyfmt](https://github.com/penelopezone/rubyfmt/issues/154) that initiated the conversation.

In truth, I only adopted trailing commas when I first published standard at @penelopezone's insistence and after a year or two of living with them, my frustration with them did not alleviate, so I was relieved by her reversal on that stance.

There are two reasons why I prefer `no_comma`:

1. Standard is designed to specify a code style that is natural and obvious to write. The way I write Ruby code at this point is standards-compliant; it's very rare for the formatter to change anything on save for me at this point. We could only accomplish this by carefully selecting a configuration that wasn't made much easier by delegating to the autofixer (it's why we don't have any sort of alignment rules like aligning `=` or `:` characters in a multiline literal or expression, because later statements could change the formatting of an earlier line). While it's certainly possible to get in the habit of always tacking on a `,` to the end of every line, this was the only rule in the whole standard set that I couldn't train myself to adopt—probably because of a lifetime of only adding a `,` to prose when I had something to add afterward. Finishing an expression with a comma feels unnatural for that reason, I believe.

2. The git churn issue strikes me as less compelling as more and more tooling will visually highlight actual changes and deemphasize unchanged text. For example, from this very PR it's not distracting at all to see this line change is nothing more than a comma change:

Screenshot 2022-08-31 at 1 25 05 PM

The reason we switched the rule instead of disabling it is because this is a major decision point that affects code readability pretty significantly and I felt it was important that Standard make a call one way or another. When you read the full `base.yml`, you'll see we have been pretty sparing in which cops we have chosen, but this was one case where encountering a codebase that mixed-and-matched styles would feel inconsistent and messy.

I realize a lot of this is subjective and being on the other side of the argument makes anything I say feel arbitrary. I think this is one issue is at least partly motivated by taste and aesthetics.

casperisfine:
No worries, and thank you for the extra context I missed.

I off course totally disagree for various reason, but as you said: taste and aesthetics 😄

searls:
Thanks.

A word I didn't use in my description above was "consistency". One of my two internal mottos with Standard is "be nothing if not consistent", and that meant that once trailing commas started causing some real headaches for rubyfmt around parameter lists and some head-scratchers for data literals in Standard, the fact that no-trailing-commas could be confidently applied safely 100% of the time, it became appealing because it was the only bulletproof consistent rule we could apply in all cases.

penelopezone:
fwiw rubyfmt has some objectively horrible syntax choices because it's the only way to be consistent. We chose this path and are sticking to it

krystof-k:
We've just started using Standard Ruby, but this is definitely a reason to ditch it. Probably one of the most important cases for a linter is to enforce trailing commas to keep the Git history clean. I strongly disagree with the aesthetics argument, this is code, not art.

fables-tales:
I assure you we thought about git diffs at the time and decided it wasn't important enough to override this decision :)

PR #458, opened 2022-09-13
Fix autocorrect fixture so disable comments are preserved

patrickbrown-dev:
Based on my understanding of 8e37104, this test was accurately failing because we want these disable comments to be preserved after autocorrect.

### DIFF
--- a/test/fixture/cli/autocorrectable-bad.rb
+++ b/test/fixture/cli/autocorrectable-bad.rb
@@ -1,7 +1,5 @@
 #Some code that violates lots of rules
 
-require "json" # standard:disable Lint/AssignmentInCondition
-
 STUFF = [
   1,
           3,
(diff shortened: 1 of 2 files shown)


searls:
This was added in https://github.com/testdouble/standard/commit/fa2ccfad69586a0d26e24e8d7b4b09c3c9bf4c9e in order to ensure that PR had a test around it. Since we are disabling the feature (as opposed to enabling a cop that, for some silly reason, _ensures_ irrelevant disable directives remain in the code), I think the right thing to do would be to revert that testing commit entirely and leave the behavior unspecified.

Propagating the comment would IMO be an example of "overspecifying" a test: adding (or keeping) assertions that don't (or any longer) specify intended behavior of the system. By locking in an incidental behavior, a test loses its power to express the author's intent, because a reader wouldn't be able to know whether a particular assertion was an expression of desired behavior or an incidental behavior that was merely being locked in

patrickbrown-dev:
Sounds good @searls! I'll modify this PR accordingly.

patrickbrown-dev:
@searls PR updated to revert fa2ccfa!

PR #459, opened 2022-09-15
Support running Standard on older Ruby project

mrbiggred:
Updated Standard to better support linting older Ruby code bases.  Standard still requires Ruby 2.6 to run but this update adds better support for Rubocop's built in [TargetRubyVersion](https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version) feature so older code bases can be linted correctly.

Please note that the the TargetRubyVersion feature only supports Ruby 2.0 and onwards.  I couldn't find this documented anywhere but did find it in the code [here](https://github.com/rubocop/rubocop/blob/master/lib/rubocop/target_ruby.rb):

```ruby
module RuboCop
  # The kind of Ruby that code inspected by RuboCop is written in.
  # @api private
  class TargetRuby
    KNOWN_RUBIES = [2.0, 2.1, 2.2, 2.3, 2.4, 2.5, 2.6, 2.7, 3.0, 3.1, 3.2].freeze
```

Rubocop did remove support for linting older Ruby versions but re-added that support in Rubocop [v1.30.0](https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#1300-2022-05-26).

This PR makes the following changes:

- Updates the `config/*.yml` files with the TargetRubyVersion.
- Adds Ruby 2.0 and 2.1 `config/` files.
- Fixes and refactors the code in the method that determines TargetRubyVersion to use.

Let me know if you have any questions or feedback on the PR.  I am looking forward to being able to use Standard on legacy code bases we are modernizing.


lib/standard/creates_config_store/sets_target_ruby_version.rb
```diff
@@ -2,20 +2,27 @@ class Standard::CreatesConfigStore
...
-      if desired_version < rubocop_supported_version
-        rubocop_supported_version
+      # This is minimum version that Rubocop can parse, not the minimum
+      # version it can run on (e.g. TargetRubyVersion).  See the following
+      # for more details:
+      #
+      # https://docs.rubocop.org/rubocop/configuration.html#setting-the-target-ruby-version
+      #
+      # https://github.com/rubocop/rubocop/blob/master/lib/rubocop/target_ruby.rb
+      min_target_ruby_version = Gem::Version.new("2.0")
```
searls, line 21: Assuming this is a magical number -- is there a better place for it to live in the repo for when we need to update it?

lib/standard/creates_config_store/assigns_rubocop_yaml.rb
```diff
@@ -18,6 +18,10 @@ def rubocop_yaml_path(desired_version)
         "ruby-1.8.yml"
       elsif desired_version < Gem::Version.new("2.0")
         "ruby-1.9.yml"
+      elsif desired_version < Gem::Version.new("2.1")
```
searls, line 21: If the minimum parseable version for Rubocop is 2.0, does this mean we need to delete tacit support for 1.8 and 1.9 code from Standard at this point?

.gitignore
```diff
@@ -7,3 +7,4 @@
 /spec/reports/
 /tmp/
 /vendor/
+/.idea/
```
searls, line 10: If you don't mind, please keep tool-specific ignores in your global excludes file https://git-scm.com/docs/gitignore

searls (changes requested):

lib/standard/creates_config_store/sets_target_ruby_version.rb
mrbiggred: I'll find a better place for this magic number.

lib/standard/creates_config_store/assigns_rubocop_yaml.rb
mrbiggred: I would leave the `ruby-1.8.yml` and `ruby-1.9.yml` files.  That way custom Ruby 1.8 and 1.9 rules can be added to those files as needed.  Rubycop might not support linting Ruby 1.8 or 1.9 but Standard can.

.gitignore
mrbiggred: Can do.

lib/standard/creates_config_store/sets_target_ruby_version.rb
mrbiggred: Added as a constant in the `sets_target_ruby_version` class.  If you have a better place for it please let me know.

.gitignore
mrbiggred: JetBrains `.idea` folder has been removed.

searls:
Thanks! So will this change have the potential to negatively impact anyone running Standard against an old Ruby today?

@mrbiggred would you mind adding trailing newlines to each file? It looks like your editor stripped several.

Before merge, @camilopayan could you take a look at this?

mrbiggred:
@searls I've added blank lines to the config files.  Let me know if any file is still missing a blank line at the end.

I don't thing these changes will negatively impact running Standard against old Ruby version.  Linting legacy legacy Ruby versions should now be more actuate.  My guess is some users will be able to remove custom ignore rules or files from the `.standard.yml`.  They might also be able to remove items from the `.standard_todo.yml` file.

If my understanding of the code is correct then Standard will default to the version of Ruby running standard unless it's overridden by the `ruby_version` setting in `.standard.yml` as shown below. 

```ruby
# standard/loads_yaml_config.rb

def construct_config(yaml_path, standard_yaml, todo_path, todo_yaml)
    {
      ruby_version: normalized_ruby_version(standard_yaml["ruby_version"]),
      
      # Other settings we aren't interested in.
    }
end

def normalized_ruby_version(version)
  return version if version && !Gem::Version.correct?(version)

  Gem::Version.new(version || RUBY_VERSION)
end
```

That said, if you or @camilopayan, notice something I missed please let me know.

searls (approved):

searls:
Looks good! This should land in @camilopayan's next release! Thanks, Chris!

mrbiggred:
@searls happy to help.  Thank you and @camilopayan for creating Standard in the first place.  I enjoy using it.  All the best.

Issue #375, opened 2022-01-06
Unexpected changes with named arguments in 1.6.0 from Style/HashSyntax


danielmorrison:
Looks like we have this new cop and I'm not a fan, so raising the issue here: Style/HashSyntax: Omit the hash value.

I'm getting code that looks like: 
```ruby
category = "foo"
Product.where(category: category)
```

Changing to: 

```ruby
category = "foo"
Product.where(category:)
```

On the one hand, I learned that this works! Cool that I can still have things to learn! On the other hand, it was pretty unexpected to me.

Looks like [`EnforcedShorthandSyntax: never`](https://docs.rubocop.org/rubocop/1.24/cops_style.html#enforcedshorthandsyntax-never) is what I'd vote for.

searls:
Agree, this seems heavy-handed in practice. It's a neat trick but it doesn't make sense to force terseness on people, especially when it's so new. (I say this as someone who routinely supports way-since-EOL'd Rubies in my new code, so this seems like something I'll be late-to-the-party to).

@camilopayan, thoughts?

darronschall:
For other folks running into this, a temporary workaround is to disable this particular cop across your entire project. In my `.standard.yml` I added:

```
# Temporarily ignore Ruby 3.1 omit hash value syntactic sugar
# see: https://github.com/testdouble/standard/issues/375
ignore:
  - '**/*':
    - Style/HashSyntax
```

whomwah:
It also causes problems if you are using a CI matrix of Ruby versions < 3.1 in CI.

searls:
Spoke with @camilopayan who agrees this rule is an overreach. We will cut a release ASAP that turns it off

searls:
also, for the record, I'm not going to enable the rule with `EnforcedShorthandSyntax: never` because I think this is a spiffy new feature that allows for terseness in lots of useful cases (like when an object exists to shovel lots attributes from one thing to the next), and it'd be a shame to ban its use in standard projects

danielmorrison:
Sounds great to me. I'm also fine not banning spiffy new things. 

You guys are awesome. Keep up the great work!

whomwah:
Thanks for the fast turnaround 👍

searls:
Well, shit. We don't have that option. I hadn't realized this was a tack-on to the (otherwise very important) `Style/HashSyntax` cop, and there's no option to be unopinionated about this

[...]

PR #467, opened 2022-10-06
Switch to consistent shorthand hash syntax

jandudulski:
The only thing that I don't like more than a shorthand hash syntax is mixed syntax. [When standard](https://github.com/testdouble/standard/issues/375) was introducing `EnforcedShorthandSyntax: either` a `consistent` option didn't exist yet (actually `either` didn't exist too, it was introduced thanks to standard :heart: ) but now it should be no brainer to switch to this one.

```ruby
# with either

# good
{foo: foo, bar: bar}

# good
{foo:, bar: bar} # !!!

# good
{foo:, bar:}
```

```ruby
# with consistent

# bad
{foo: , bar: bar}

# good
{foo:, bar:}

# bad
{foo: , bar: baz}

# good
{foo: foo, bar: baz}
```

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -1355,7 +1355,7 @@ Style/HashLikeCase:
 Style/HashSyntax:
   Enabled: true
   EnforcedStyle: ruby19_no_mixed_keys
-  EnforcedShorthandSyntax: either
+  EnforcedShorthandSyntax: consistent
 
 Style/HashTransformKeys:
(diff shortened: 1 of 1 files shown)


camilopayan:
I think that `either` fits well with Ruby's other syntax around method/function signatures. I'll admit the bad mixed examples looks fine to me.

Thanks for the contribution, though!

PR #475, opened 2022-10-26
LSP server

### LINKED ISSUE #472: Do you want a LSP server in this project?
Before, I was using solargraph and doing some not-great monkeypatches to get it to use standard instead of raw rubocop. But I realized all I really wanted from solargraph was just diagnostics and formatting and none of the other LSP features from solargraph. 

Over the last few days I've made a custom LSP server that just does formatting and diagnostics, and with a few tricks, like reusing the huge rubocop_config_store, it can do things nearly real-time. Close enough for me at least:

https://user-images.githubusercontent.com/1973/197632689-750d9ecf-e519-4696-9620-33c5ec52c434.mov

I wonder if this is something you'd like in the standard project itself? It could be nice to just run `standard --lsp` or something.  It's a bit under 200loc (no tests yet though) and takes one dependency to `language_server-protocol`

will:
This adds an LSP server that runs with the flag `--lsp`. It does the lsp jsonrpc over stdio. Functionality wise, this works for everything I need. I've only tested with neovim 0.8 though. It only operates on documents that you open, rather than going through your entire project looking for ruby files. 

Test coverage isn't great, but I think they cover the major parts:

* telling the lsp client (an editor) on init what the server supports and to send full documents instead of patches
* getting diagnostics
* formatting

_I'm_ okay with this level of testing to start, but I'd understand if you weren't.

I think this is a good starting point for functionality, but there some improvements left on the table:

If several changes come in while it was doing some other work, the stdin reader could collapse the changes and only take the last full change.

Editors could just send the small changes on each change, but then the server would have to be able to patch those into the copy it has in memory. Not impossible, but defiantly a lot more work. Would also make the previous collapsing improvement not necessary.

Format results could also be a series of patches rather than the entire document.

It could go through and find all the files in your workspace, but idk, I think just working on open files is good enough.

It might be possible to do the diagnostics and formatting in memory rather than bouncing through a tempfile. The solargraph moneky patch I was using before did this originally, but it stopped working and I couldn't figure out why, which lead me down to writing a small lsp form scratch in the first place.

Also please let me know what sort of git etiquette you'd like for changes on the PR. rebasing into the original commit or fixup additional patches or whatnot.

closes #472

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,6 +2,7 @@ PATH
   remote: .
   specs:
     standard (1.16.1)
+      language_server-protocol (= 3.17.0.1)
       rubocop (= 1.35.1)
       rubocop-performance (= 1.14.3)
 
@@ -13,6 +14,7 @@ GEM
     docile (1.4.0)
     gimme (0.5.0)
     json (2.6.2)
+    language_server-protocol (3.17.0.1)
     method_source (1.0.0)
     minitest (5.16.3)
     parallel (1.22.1)
--- a/lib/standard/lsp/server.rb
+++ b/lib/standard/lsp/server.rb
@@ -15,8 +15,8 @@ module Standard
 
       def initialize(standardizer)
         self.standardizer = standardizer
-        self.writer = Proto::Transport::Stdio::Writer.new
-        self.reader = Proto::Transport::Stdio::Reader.new
+        self.writer = Proto::Transport::Io::Writer.new($stdout)
+        self.reader = Proto::Transport::Io::Reader.new($stdin)
         self.logger = $stderr
         self.text_cache = {}
 
@@ -98,6 +98,7 @@ module Standard
 
         lsp_diagnostics = offenses.map do |o|
           code = o[:cop_name]
+
           msg = o[:message].delete_prefix(code)
           loc = o[:location]
 
--- a/lib/standard/lsp/server.rb
+++ b/lib/standard/lsp/server.rb
@@ -1,0 +1,140 @@
+require "language_server-protocol"
+require_relative "standardizer"
+
+module Standard
+  module LSP
+    class Server
+      Proto = LanguageServer::Protocol
+      SEV = Proto::Constant::DiagnosticSeverity
+
+      def self.start(standardizer)
+        new(standardizer).start
+      end
+
+      attr_accessor :standardizer, :writer, :reader, :logger, :text_cache, :subscribers
+
+      def initialize(standardizer)
+        self.standardizer = standardizer
+        self.writer = Proto::Transport::Stdio::Writer.new
+        self.reader = Proto::Transport::Stdio::Reader.new
+        self.logger = $stderr
+        self.text_cache = {}
+
+        self.subscribers = {
+          "initialize" => ->(request) {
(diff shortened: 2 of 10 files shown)


searls:
Thanks! Will review soon. I'd encourage you to use whatever git technique you like

searls:
I dig your code style! Thanks for tidying a few things up.

What's the bare minimum setup I could use to test this actually working for myself locally?

will:
Thanks :)

The most minimal config for neovim 0.8+ would be this, with replacing the path:


```lua
-- /tmp/init.lua
vim.opt.signcolumn = "yes" -- otherwise it bounces in and out
vim.api.nvim_create_autocmd("FileType", {
  pattern = "ruby",
  group = vim.api.nvim_create_augroup("RubyLSP", { clear = true }),
  callback = function()
    vim.lsp.start {
      name = "standard",
      cmd = { "/Users/will/code/standard/exe/standardrb", "--lsp" },
    }
  end,
})
```

Then you can edit a ruby file with it `nvim -u /tmp/init.lua a.rb`

It should auto do the in-line virtual text for diagnostics (which I find too noisy to have on all the time personally). Then to format you can do `:lua vim.lsp.buf.formatting()` which you would want to map to something like `<leader>lf` with an `on_attach` function, but it is not minimal to do that ;)

foca:
Been playing with this on a personal project and it's **fantastic** :smile::heart:

(Beyond the setup code on that vim config file, I also added an autocmd to format the code on `BufWrite` that works instantly :))

searls:
Seems to work great! Finally tested it myself

Screenshot 2022-12-03 at 11 43 57 AM

will:
I thinks you posted the wrong screenshots and that's what I appreciates about you

searls:
Looks great to me, thanks @will!

 lsp-demo

searls:
Landed in v1.19.0! Great work, @will. I'm super grateful for your initiative, your contribution, and your patience. I'm also really impressed with how well you structured your code similarly to what was already in place!

will:
Thanks for the kind words and for merging it :)

bluz71:
I was using Standard Ruby via [`null-ls`](https://github.com/jose-elias-alvarez/null-ls.nvim) inside Neovim. I just switched over from that to the new LSP option, and the difference is **incredible**.

From very slow, to near instant. Wow!

`null-ls` invokes the `standardrb` command line on each edit and the lag is very noticeable; up to 2 seconds (sometimes more).

The new `--lsp` option on the other-hand is near realtime; I never thought that was possible for anything related to Rubocop. But seeing is believing.

This feature needs to be advertised wide and far; especially integration into [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig).

From I have seen, this simple LSP addition easily competes with Shopify's [Ruby LSP](https://github.com/Shopify/ruby-lsp) which appears to be focussed on diagnostics and formatting, but not completion (similar to this new `--lsp` option).

Note, I still use Solargrapgh for code completion and navigation (whilst disabling diagnostics and formatting). Two LSP's running at the same time, with no overlap, works just fine.

Great work @will and @searls.

will:
Thanks @bluz71 it's very nice to hear that you like it :)

PR #479, opened 2022-11-10
Update rubocop to 1.37.1

CKolkey:
This patch updates rubocop to `1.37.1`. I tried updating to the latest rubocop (`1.38.0`) but there appears to be a bug in rubocop which raises an exception when running the test suite, so it should likely be avoided until `1.38.1`.


camilopayan:
All up to date! Thanks :)

PR #484, opened 2022-11-14
Set StandardRb to a monthly update cadence

camilopayan:


searls:
This makes sense to me as a first start. Sorry for the delay in looking at it. Merge it and try it out!

PR #494, opened 2022-12-16
Enable Extra Spacing Before Trailing Comments

samanthaoldenburg:
## Summary

The `AllowBeforeTrailingComments` flag for `Layout/ExtraSpacing` was explicitly disabled in https://github.com/testdouble/standard/pull/125, as part of a bulk change to add Rubocop's defaults. 

I think this should be explicitly set to `true`. It allows for better grouping of comments when they trail.

## Example of Before and After
 
**Before**

```ruby
class Student < ApplicationRecord
  has_one :counselor, class: "Faculty" # The guidance counselor helps a student pick their classes
  has_one :school # A student can only be enrolled in one school at a time
  has_many :classes # A student needs to be enrolled in classes
  has_one :homeroom, class: "Class" # A student starts each day of school in their homeroom
end
```

**After**

```ruby
class Student < ApplicationRecord
  has_one :counselor, class: "Faculty"   # The guidance counselor helps a student pick their classes
  has_one :school                        # A student can only be enrolled in one school at a time
  has_many :classes                      # A student needs to be enrolled in classes
  has_one :homeroom, class: "Class"      # A student starts each day of school in their homeroom
end
```

I think this leads to more readable code when using trailing commas.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -182,7 +182,7 @@ Layout/EndOfLine:
 Layout/ExtraSpacing:
   Enabled: true
   AllowForAlignment: false
-  AllowBeforeTrailingComments: false
+  AllowBeforeTrailingComments: true
   ForceEqualSignAlignment: false
 
 Layout/FirstArgumentIndentation:


Kuraiou:
This recommended change leads to more readable code any time there is a set of lines illustrating a process, irrelevant of comma usage.

unlike other spacing issues, aligned comments are a common tool for tutorial and demo code (walking through steps of a process), and the inclusion of extra whitespace before a comment does not hurt real code legibility.

it is, from my point of view, _exceedingly_ unlikely that an individual will be grepping through their codebase just for trailing commas (e.g. searching for `"[a-zA-Z] #"`). It is exceedingly unlikely that this change will negatively impact any machine reading tools as well, since trailing comments are not relevant to the executing code's point of view.

I see no value in maintaining the rubocop default value for this cop and i see value in setting it to `true`. I would love to hear arguments otherwise.

saturnflyer:
I strongly agree with this. Spacing before comments should be allowed but alignment not enforced.

searls:
Don't see anything disagreeable here. It definitely seems like this rule was set without intention on account of locking in the rubocop defaults explicitly

PR #506, opened 2023-01-14
Add a new feature to Standard to allow it to run plugins & custom cops

searls:
This one has been a long time coming. Right now, if you run Standard but have any cause to run any additional RuboCop rules (whether a gem plugin like [betterlint](https://github.com/Betterment/betterlint) for your Rails app) or a custom cop that's maintained in your own codebase, you've had to either run `standardrb` alongside `rubocop` or load Standard's ruleset into a `rubocop` configuration like [this guide by Evil Martians](https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard) explains.

Neither option is great—the former is slow and the latter risks overwriting or modifying Standard's rules—often invisibly—such that code you thought was Standard-compliant was actually being impacted by an arbitrary gem you required.

To remedy this issue, this PR adds a new configuration property for your `.standard.yml` file called `extend_config`. It takes one or more _RuboCop_ YAML file locations which can require files and gems of other cops, configure their rules, and even inherit from yet other RuboCop configurations.

To wit, you could add this to your `.standard.yml` file:

```yml
extend_config:
  - .betterlint.yml
```

And `.betterlint.yml` might look like:

```yml
require:
  - rubocop/cop/betterment.rb

AllCops:
  DisabledByDefault: true

Betterment/UnscopedFind:
  Enabled: true

  unauthenticated_models:
    - SystemConfiguration
```

From then on, any invocation of `standardrb` or `rake standard` will run your code through _both_ Standard's ruleset as well as `Betterment/UnscopedFind`.

At runtime, Standard will merge in all the new rules while excluding any changes to the built-in cops it cares about (which is to say, any cops that ship with either the `rubocop` or `rubocop-performance` gems). It will also allow the majority of top-level configurations on the `AllCops` section to be modified (since they won't affect Standard's core behavior).

Hopefully this makes Standard easier to adopt and to use exclusively for your Ruby linting & formatting needs! 💚

Todo list:

- [x] Make it work in 2.7 lol

### DIFF
--- a/README.md
+++ b/README.md
@@ -492,12 +492,50 @@ either:
   gem to your build, which will automatically ensure that the comment is
   prepended for every applicable file in your project
 
-## How do I use Standard with RuboCop extensions?
+## How do I use Standard with RuboCop extensions or custom rules?
 
-This is not officially supported by Standard. However, Evil Martians wrote up [a
-regularly updated
+If you want to use Standard in conjunction with RuboCop extensions or custom
+cops, you can specify them in your own [RuboCop configuration
+YAML](https://docs.rubocop.org/rubocop/configuration.html) files and
+`.standard.yml` using the "extend_config:` setting:
+
+```yml
+# .standard.yml
+extend_config:
+  - .betterlint.yml
+```
+
+Which in turn would refer to a RuboCop configuration file:
+
+```yml
+# .betterlint.yml
+require:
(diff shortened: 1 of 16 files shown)


searls:
Landed in v1.22.0 🎉

PR #508, opened 2023-01-16
Allow empty else that have comments

MaxLap:
This option was added in Rubocop 1.32 https://github.com/rubocop/rubocop/blob/master/CHANGELOG.md#1320-2022-07-21.
It mirrors what Standard already allows for EmptyWhen.


searls:
Thanks for spotting this!

PR #511, opened 2023-01-17
Add Arrows to list of standard users

swanson:
We are happy users of `standard` for our Rails app at https://arrows.to


searls:
Yay! Thanks Matt!

PR #523, opened 2023-02-11
What if I told you there were more LSP improvements

searls:
Using stdin instead of tempfiles is a significant speed improvement.

Running before this commit: 0.129s - 0.141s
Running lsp_test.rb after this commit: 0.075s - 0.085s

- [x] use stdin instead of temp files
- [x] send a cwd-relative path name correctly to rubocop
- [x] figure out how to ignore paths when .standard.yml would ignore them (possibly using `TargetFiles#find` and passing the basename of the file's dir to see if it comes back? Would result in false positive for unsaved files…)

### DIFF
--- a/lib/standard/lsp/logger.rb
+++ b/lib/standard/lsp/logger.rb
@@ -1,8 +1,18 @@
 module Standard
   module Lsp
     class Logger
+      def initialize
+        @puts_onces = []
+      end
+
       def puts(message)
         warn("[server] #{message}")
+      end
+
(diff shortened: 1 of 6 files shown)


searls:
For anyone playing "Weekend Project: Home Edition", the `force_exclusion: true` CLI option made ignores basically work automatically so long as I took the `file://` URI off the path. Not only does it require almost no extra code for computing paths (phew), but it seems to work in sample projects straight away

PR #541, opened 2023-03-27
Added Spinal (https://spinalcms.com/) as a Ruby Standard-compliant team

eelcoj:
Standard helps keep the Spinal code-base consistent (and helps devs sometimes learn something new). 🙌


searls:
Thanks for the note!

PR #543, opened 2023-03-31
Return an ExecuteCommandOptions struct in place of a boolean

### LINKED ISSUE #542: LSP server fails to initialise in Helix editor
When using `standardrb --lsp` in Helix I get error messages in the log:

```
2023-03-31T18:05:48.225 helix_lsp [ERROR] failed to initialize language server: failed to parse: invalid type: boolean `true`, expected struct ExecuteCommandOptions
```

As this worked in Neovim, I initially reported the bug to Helix: https://github.com/helix-editor/helix/issues/6506 who responded:

> This is a bug in StandardRB: it's sending "executeCommandProvider":true in the initialization response but that field can't be a boolean in the spec: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#executeCommandOptions

benjamineskola:
Return an ExecuteCommandOptions struct for `execute_command_options`, rather than a boolean

Fixes #542

This at least seems to allow Helix to register the LSP server and I get diagnostics through from StandardRB, whereas previously nothing at all was working. Formatting doesn't seem to work but that may well be a configuration issue at my end, I'm new to Helix.

### DIFF
--- a/lib/standard/lsp/routes.rb
+++ b/lib/standard/lsp/routes.rb
@@ -28,7 +28,9 @@ module Standard
           capabilities: Proto::Interface::ServerCapabilities.new(
             document_formatting_provider: true,
             diagnostic_provider: true,
-            execute_command_provider: true,
+            execute_command_provider: Proto::Interface::ExecuteCommandOptions.new(
+              commands: ["standardRuby.formatAutoFixes"]
+            ),
             text_document_sync: Proto::Interface::TextDocumentSyncOptions.new(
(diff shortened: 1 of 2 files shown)


searls:
Perfect. Thanks for saving me the time

searls:
Landed in 1.25.4

searls:
Annnd had to unwind this (I just deleted the capability entirely) because this patch broke the VS Code extension, presumably because it registers a command of the same name on the client side https://github.com/testdouble/standard/issues/544

PR #547, opened 2023-04-03
Add configs for Ruby 3.1 and 3.2

jandudulski:


searls:
Did you submit this PR because of a failure you encountered?

We only bundle Ruby-version-specific YAML files when we need to configure the rules differently for an old Ruby (most often, because it's fallen out of RuboCop's support and a specific rule needs to be manually disabled for versions beneath its lowest target ruby version), which shouldn't be the case here.

jandudulski:
hi @searls, mainly I did this for `TargetRubyVersion` option. Should I just overwrite it in my `.rubocop.yml` config?

PR #560, opened 2023-05-10
🔧 config/ruby-3.2.yml

pboling:
- Fixes https://github.com/standardrb/standard/issues/559
- Watched https://www.youtube.com/watch?v=8rLe_qmH84k - add latest Ruby
- Note: *current* release of `standard` already depends on the released version of RuboCop (1.50.2) that [added support](https://github.com/rubocop/rubocop/commit/5fa1573ef01ef9f93808a7a249e53f2930d3002a) (experimental) for current `ruby-head` (which is Ruby 3.3) via `TargetRubyVersion: 3.3`.
- There was one failing spec because the test expects Ruby 3.2 to be the base config, instead of having the base represent Ruby 3.3 (current ruby-head).
- Latest rubygems/bundler was [released 5 hours](https://github.com/rubygems/rubygems/releases/tag/v3.4.13) ago!

### DIFF
--- a/lib/standard/base/plugin.rb
+++ b/lib/standard/base/plugin.rb
@@ -55,6 +55,8 @@ module Standard::Base
         "ruby-3.0.yml"
       elsif desired_version < Gem::Version.new("3.2")
         "ruby-3.1.yml"
+      elsif desired_version < Gem::Version.new("3.3")
+        "ruby-3.2.yml"
       else
         default
       end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -39,7 +39,7 @@ GEM
       rubocop-ast (>= 1.28.0, < 2.0)
       ruby-progressbar (~> 1.7)
       unicode-display_width (>= 2.4.0, < 3.0)
-    rubocop-ast (1.28.0)
+    rubocop-ast (1.28.1)
       parser (>= 3.2.1.0)
     rubocop-performance (1.16.0)
       rubocop (>= 1.7.0, < 2.0)
@@ -72,4 +72,4 @@ DEPENDENCIES
   standard!
 
 BUNDLED WITH
-   2.3.26
+   2.4.13
--- a/config/ruby-3.1.yml
+++ b/config/ruby-3.1.yml
@@ -1,4 +1,4 @@
-inherit_from: ./base.yml
+inherit_from: ./ruby-3.2.yml
 
 AllCops:
   TargetRubyVersion: 3.1
--- a/config/ruby-3.2.yml
+++ b/config/ruby-3.2.yml
@@ -1,0 +1,4 @@
+inherit_from: ./base.yml
+
+AllCops:
+  TargetRubyVersion: 3.2
--- a/test/standard/base/plugin_test.rb
+++ b/test/standard/base/plugin_test.rb
@@ -8,7 +8,8 @@ module Standard::Base
     end
 
     def test_paths
-      assert_match "base.yml", @subject.rules(LintRoller::Context.new(target_ruby_version: "3.2.1")).value.to_s
+      assert_match "base.yml", @subject.rules(LintRoller::Context.new(target_ruby_version: "3.3.0")).value.to_s
+      assert_match "ruby-3.2.yml", @subject.rules(LintRoller::Context.new(target_ruby_version: "3.2.1")).value.to_s
       assert_match "ruby-2.7.yml", @subject.rules(LintRoller::Context.new(target_ruby_version: Gem::Version.new("2.8.2"))).value.to_s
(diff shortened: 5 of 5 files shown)


lib/standard/base/plugin.rb
```diff
@@ -55,6 +55,8 @@ def determine_yaml_path(desired_version)
         "ruby-3.0.yml"
       elsif desired_version < Gem::Version.new("3.2")
         "ruby-3.1.yml"
+      elsif desired_version.segments[0] == 3 && desired_version.segments[1] == 2 # exactly 3.2.x so HEAD will fall through to default.
```
searls, line 58: Any reason not to follow the pattern above like `desired_version < Gem::Version.new("3.3")`?

searls (changes requested):

lib/standard/base/plugin.rb
pboling: Excellent question.  I remember thinking that doing it the same way wouldn't work, but I can see now that was entirely a brain fart.  Fixing!

pboling: Fixed (& rebuilt the commit history)

searls:
Could you help me understand what this PR is fixing? You mentioned a failing spec, but after [adding `head` to our test workflow](https://github.com/standardrb/standard/actions/runs/4975416801) it passed without issue.

In general, we have only added version-specific YAML files when there is a Standard-enabled rule that is pegged to a specific version (in terms of language compatibility), and this PR doesn't specify any rules, so it seems like it'd effectively be a no-op

pboling:
I apologize, it seems there has been a complete miscommunication about my motivation for this PR as described in this issue: https://github.com/standardrb/standard/issues/559

Apologies for 1) burying the lede, 2) represeting the concern so poorly, and 3) starting out with misleading background information.

I am not using standard's runtime.  I am using the alternative loading method where you can load the specific Ruby version explicitly from Rubocop.  You link to [Evil Martian's writeup](https://evilmartians.com/chronicles/rubocoping-with-legacy-bring-your-ruby-code-up-to-standard) about this method from your readme.  Except I take that one step further by using [Evil Martian's _other_ guide](https://dev.to/evilmartians/rubocoping-with-legacy-gme) on setting up `standard`, `rubocop`, and `rubocop-gradual` (a new-ish tool by Evil Martian's).  This approach is, IMO, the "best-possible-setup" because of the way `rubocop-gradual` works.  Seeing it work now across many projects, I don't think I can go back to `rubocop_todo.yml`.

In the scenario where I have many dozens of libraries and apps I am intending to migrate to this pattern, many of them forever targeting discrete, specific, versions of Ruby (e.g. [`rubocop-lts` suite](https://rubocop-lts.gitlab.io/) of gems), I am interested in having a versioned file that can be a target.

In case any rules are added in the future to RuboCop that break on an older version of Ruby (3.2 will eventually be old!), they'd need to be disabled.  Creating the file in the future would mean updating all the gems that had been targeting "base.yml", which I may not have time to do.  Creating the file now means that libraries can depend on `standard`, and as standard updates these Ruby version files to continue working with each release of RuboCop, gems and apps depending on Standard will just keep working!  Standard becomes a clearing house for maintaining the versioned Ruby x Cop matrix, which RuboCop itself has decided not to do (they have enough on their plate).

I love Standard, and it will be immensely more useful for the niche of library that I build if it always includes a versioned file for every minor version of Ruby.

pboling:
Regarding the failing spec... I left out the implied part.  The test broke because of the changes made (3.2 != 3.3); it isn't fixing an existing broken test.

Perhaps I should have said:

> "[With these changes in place] there was one failing spec _because_ the test expects Ruby 3.2 to be the base config, instead of having the base represent Ruby 3.3 (current ruby-head), [and with these changes base is changed to be Ruby 3.3]".

pboling:
> this PR doesn't specify any rules, so it seems like it'd effectively be a no-op

How do we account for these then?

- https://github.com/standardrb/standard/blob/main/config/ruby-2.0.yml
- https://github.com/standardrb/standard/blob/main/config/ruby-2.1.yml

The only rules they add are important, and `ruby-3.2.yml` in this PR adds _all_ of the same ones.

```
inherit_from: ./ruby-2.1.yml

AllCops:
  TargetRubyVersion: 2.0
```

```
inherit_from: ./ruby-2.2.yml

AllCops:
  TargetRubyVersion: 2.1
```

Compare that with:

```
inherit_from: ./base.yml

AllCops:
  TargetRubyVersion: 3.2
```

AFACT this is aligned with the ethos of this repository.

searls:
Merged these changes (Bundler update aside) and released in standard@1.28.3

PR #563, opened 2023-05-29
🐛 Ruby 1.8 & 1.9 do not support percent literal syntax

pboling:
Standard's `base.yml` has the following rule, which is great:
```
# Set in standard's base.yml.
Style/PercentLiteralDelimiters:
  Enabled: true
  PreferredDelimiters:
    default: ()
    '%i': '[]'
    '%I': '[]'
    '%r': '{}'
    '%w': '[]'
    '%W': '[]'
```

However, it needs to be disabled for the oldest Rubies.  The percent literal syntax is not supported by codebases that need to continue to support Ruby 1.8 and 1.9.  The answer to that in the RuboCop documentation is ([from `Style/SymbolArray`](https://msp-greg.github.io/rubocop/RuboCop/Cop/Style/SymbolArray.html)):

> Checks for array literals made up of symbols that are not using the %i() syntax.
> Alternatively, it checks for symbol arrays using the %i() syntax on projects which do not want to use that syntax, perhaps because they support a version of Ruby lower than 2.0.

This means that the enforcement can be set to ensure compatibility with Ruby 1.8 and 1.9, but within Standard's rules we will merely disable it.

```
Style/PercentLiteralDelimiters:
  Enabled: false
```

Fixes https://github.com/standardrb/standard/issues/562

Note: since standard's goal is not to prioritize enforcement of compatibility with older versions of Ruby, I've added the "compatible" version of these changes to [`standard-rubocop-lts`](https://github.com/rubocop-lts/standard-rubocop-lts) as a "buffer" against unexpectedly breaking old rubies (see also: https://github.com/standardrb/standard/issues/561#issuecomment-1563291835).

### DIFF
--- a/config/ruby-1.9.yml
+++ b/config/ruby-1.9.yml
@@ -5,3 +5,7 @@ Style/Encoding:
 
 Style/HashConversion:
   Enabled: false
+
+# Percent delimiters are not supported until Ruby 2.0
+Style/PercentLiteralDelimiters:
+  Enabled: false


config/ruby-1.9.yml
```diff
@@ -5,3 +5,13 @@ Style/Encoding:
...
+
+# Percent delimiters are not supported until Ruby 2.0
+Style/PercentLiteralDelimiters:
+  Enabled: false
+
+Style/SymbolArray:
+  EnforcedStyle: brackets
+
+Style/WordArray:
+  EnforcedStyle: brackets
```
searls, line 17:
```suggestion
```

These rules are not enabled, so please remove them here.

searls (changes requested):

config/ruby-1.9.yml
pboling: It seems we miscommunicate a lot.  If you have any ideas on how I can improve my issue reports I'm all ears.  I will attempt to restate the issue, but your reason for asking me to remove them **is the reason I can't**.

pboling:
OK, to set the stage a bit the following three rules are extremely related, however they are not equivalent, they do not contradict each other, nor do they step on each other.
```
# 👇 enabled in base.yml; but is *incompatible* with Ruby < 2.0
Style/PercentLiteralDelimiters

# 👇 untouched by standard's configs,
# however they have a setting that supports Ruby < 2.0, which is brackets
Style/SymbolArray
Style/WordArray
```





<details>
  <summary>As an aside...</summary>
As an aside, the alternative to the `brackets` setting is `percent`, and given the config you have in `base.yml` for `Style/PercentLiteralDelimiters`, it might make sense to also set `percent` for `Style/SymbolArray` and `Style/WordArray`.
</details>

pboling:
Thanks for the explanation on ruby.social @searls - the approach makes sense now.  I'll update this PR to simply disable the percent literal rule.

config/ruby-1.9.yml
pboling: @searls Fixed!

searls:
Perfect, thank you for understanding

PR #564, opened 2023-05-29
Ignore workspace/didChangeConfiguration

ryansch:
When using `standard` with [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig), the server receives a `workspace/didChangeConfiguration` message immediately after starting.

Standard currently returns an `INVALID_SERVER_MESSAGE` which causes neovim to display lots of angry errors.

I've added a route to ignore this LSP message while logging that we're doing that.

### DIFF
--- a/lib/standard/lsp/routes.rb
+++ b/lib/standard/lsp/routes.rb
@@ -75,6 +75,10 @@ module Standard
         @writer.write({id: request[:id], result: format_file(uri)})
       end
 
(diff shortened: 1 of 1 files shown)


searls:
Seems reasonable. I'm 99% sure my VS Code extension is coded to restart the server when the configuration changes, so it doesn't immediately make sense for the LSP to do anything more fancy for its sake, either

PR #572, opened 2023-07-13
Update Rubocop CLI documentation

mjgiarlo:
TIL that `standard-custom` and `standard-performance` may  need to be added to the `require` list within the Rubocop configuration (thx, @searls!). Adding this so no one else trips on this and bogs down the issue tracker about an unsupported pattern.

### DIFF
--- a/README.md
+++ b/README.md
@@ -450,12 +450,14 @@ default rulesets, just as you would any other gem:
 ```yaml
 require:
   - standard
+  - standard-custom
+  - standard-performance
   - rubocop-performance
 
 inherit_gem:
   standard: config/base.yml
-  standard-performance: config/base.yml
   standard-custom: config/base.yml
+  standard-performance: config/base.yml
 ```
 
 ## Who uses Standard Ruby?


README.md
```diff
@@ -450,12 +450,14 @@ default rulesets, just as you would any other gem:
...
   - standard
+  - standard-custom
+  - standard-performance
   - rubocop-performance
 
 inherit_gem:
   standard: config/base.yml
-  standard-performance: config/base.yml
   standard-custom: config/base.yml
+  standard-performance: config/base.yml
```
mjgiarlo, line 460: Putting these in alpha order per above. 🤷🏻‍♂️

searls:
You can probably remove rubocop-performance in this case since standard-performance owns that dep and will require it

searls:
You can probably remove rubocop-performance in this case since standard-performance owns that dep and will require it

mjgiarlo:
@searls 💬 
> You can probably remove rubocop-performance in this case since standard-performance owns that dep and will require it

Alas, when `- rubocop-performance` is removed from the list:

```shell
$ bundle exec rubocop
Error: `Performance` cops have been extracted to the `rubocop-performance` gem.
(obsolete configuration found in /media/mjg/44aee93d-a48a-4ad3-ae90-4b084bb04b69/rvm/gems/ruby-3.2.2@h2/gems/standard-performance-1.1.1/config/base.yml, please update it)
```

mjgiarlo:
@searls The method in `standard-performance` that does the Rubocop tricking is never invoked when using the Rubocop CLI. **I'm not advocating for a change like this**, but I found that slapping this code in `lib/standard/performance.rb` is one way to avoid needing `rubocop-performance` in the Rubocop's `require` list:

```ruby
module Standard
  module Performance
    Plugin.new({}).rules(
      Standard::Plugin::CreatesRunnerContext.new.call(
        Standard::LoadsYamlConfig.new.call(
          ResolvesYamlOption.new.call(ARGV, Dir.pwd, "--config", ".standard.yml"),
          ResolvesYamlOption.new.call(ARGV, Dir.pwd, "--todo", ".standard_todo.yml")
        )
      )
    )
  end
end
```

Unless you think this is ripe for further work or analysis, maybe the README change in this PR is good enough for now?

searls:
> @searls The method in standard-performance that does the Rubocop tricking is never invoked when using the Rubocop CLI. I'm not advocating for a change like this, but I found that slapping this code in lib/standard/performance.rb is one way to avoid needing rubocop-performance in the Rubocop's require list:

Thanks. Yeah, I definitely would not add a circular dependency to standard-performance onto standard over this (which is still an unsupported use, so YMMV)

PR #582, opened 2023-09-18
Adds pre-commit hooks support

epetkus:
Adds support for [pre-commit](https://pre-commit.com/) hooks management framework.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,6 @@
+- id: standardrb
+  name: Standard Ruby
+  entry: standardrb
+  language: ruby
+  types: ["ruby"]
+  args: ["--fix"]


epetkus:
@searls Would it be possible to get your eyes on this? Having an "official" support for `standardrb` with a pre-commit hooks framework would be nice. Thanks!

searls:
I'm sorry, but I'm :-1: on this practice as well as this implementation.

* Team-wide enforcement of pre-commit hooks that can be used to prevent commits until issues are resolved will, in practice, often have the effect of delaying people from "saving their work," elongating feedback loops (which we should strive to make fast) and delaying the integration of code (which we should strive to make continuous). I've seen them used extensively in a number of environments over the years, and in a majority of cases they were instituted by—hate to be frank here—tech leaders with an authoritarian bent and imposed on others who weren't given a choice in the matter but to run them (robbing them of a sense of agency and autonomy over how they work).
* Open source "meta tools" like [pre-commit](https://pre-commit.com) typified by requiring every other tool on the planet to commit a configuration file are, in my opinion, fundamentally flawed in their design. If I were building a tool like pre-commit, I wouldn't  bottleneck my tool's adoption around whether users are successful in lobbying or pressuring other open sources to shove my specialized configuration file into their dotfile junk drawer. Instead, I'd design a community-managed repository as a fallback (like [tapioca does with Sorbet typings](https://github.com/Shopify/rbi-central)) or a decentralized way for users to author and share their own configurations that target specific tools

For these reasons, I'm personally opposed to tools of this sort in the absence of an overwhelmingly clear value proposition.

PR #584, opened 2023-10-09
Fix order of arguments in `Standard::Base` test assertions

jackjennings:
This is a minor fix to these tests to make failing test output easier to diagnose. Currently when these tests fail the "actual" and "expected" output are inverted.

I noticed this while working on a lint roller plugin that was a fork of `standard-custom` (where this issue also exists).


searls:
Right on, thanks

PR #587, opened 2023-10-18
Fix generating todo when todo file already exists

### LINKED ISSUE #565: `standardrb --generate-todo` replaces any existing todo file with deltas only
@mrbiggred FYI,

I was introducing `standard-rails` to the project I'm trying to convert from plain rubocop (because @searls fixed standardrb/standard-rails/issues/7, tyvm), but when I ran `standardrb --generate-todo` it replaced the existing todo file with one that only contained ignores for the Rails cops, losing the previous ignores.

The fix was to remove .standard_todo.yml first, but I wonder if that should be documented or if it can be automated.

mrbiggred:
Fixes #565 and should allow PR #176 to be closed.  To duplicate the bug:

1) Run `standardrb --generate-todo` on a project with some linting errors.
2) Check the `.standard_todo.yml` file to see what linting rules are ignored.  For example:

```yml
ignore:
- lib/standard.rb:
  - Layout/IndentationConsistency
```

3) Run `standardrb --generate-todo` a second time.  Without this fix it would not list the existing linting rules in the old `.standard_todo.yml` file.  for example, the new file would NOT contain:

```yml
ignore:
- lib/standard.rb:
  - Layout/IndentationConsistency
```

The fix was to skip the loading of an existing `.standard_todo.yml` file generating the new one.  Let me know what you think and if you have any feedback to improve this PR.


searls:
LGTM. Sorry for missing this prior

mrbiggred:
Wow, that was fast.  Thank for you for merging the fix and ping me if there are any issues now that it's merged.

PR #600, opened 2024-01-12
Use consistent style for `Lint/SymbolConversion`

aliismayilov:
I'd like to suggest to use consistent style for `Lint/SymbolConversion`. Examples from [rubocop docs](https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Lint/SymbolConversion):
```ruby
# bad
{
  a: 1,
  'b-c': 2
}

# good (quote all keys if any need quoting)
{
  'a': 1,
  'b-c': 2
}

# good (no quoting required)
{
  a: 1,
  b: 2
}
```

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -773,6 +773,7 @@ Lint/SuppressedException:
 
 Lint/SymbolConversion:
   Enabled: true
+  EnforcedStyle: consistent
 
 Lint/Syntax:
   Enabled: true


searls:
This is a tricky one, because there are two competing interests here. On one hand, I agree that consistency in hash format is important. For example, if a single `=>` is required in a hash literal, Standard converts all entries to use `=>` so that it and `:` aren't interspersed. 

On the other hand, enforcing that the presence of a single illegal identifier should force _all_ such identifiers to be quoted seems overkill. For example, if you carried the logic that the presence of one illegal symbol literal (`:b-c`) should require all legal symbol literals to be quoted, you could make the same argument for illegal method names (or any other construct).

I'd definitely disagree with a rule that did this, for example:

```ruby
# bad
class Foo
  def legal_name
  end
  
  define_method :'illegal-name' do
  end
end

# good
class Foo
  define_method :legal_name do
  end
  
  define_method :'illegal-name' do
  end
end
```

One additional argument against, is that I'm always cautious of rules where a later line could force changes in prior lines. So if the 5th line of a hash literal contains an illegal symbol literal name, it shouldn't cause git churn for lines 1-4 without good reason. I think the `=>`/`:` rises to that bar, but I don't think this does.

As a result, I'd be :-1: to making this change.

searls:
@luizkowalski no it doesn't. I was using it to illustrate the broader point of why I don't think consistency is an unqualified principle in cases where you're working around illegal identifiers.

aliismayilov:
> This is a tricky one, because there are two competing interests here. On one hand, I agree that consistency in hash format is important. For example, if a single `=>` is required in a hash literal, Standard converts all entries to use `=>` so that it and `:` aren't interspersed.
> 
> On the other hand, enforcing that the presence of a single illegal identifier should force _all_ such identifiers to be quoted seems overkill. For example, if you carried the logic that the presence of one illegal symbol literal (`:b-c`) should require all legal symbol literals to be quoted, you could make the same argument for illegal method names (or any other construct).
> 
> I'd definitely disagree with a rule that did this, for example:
> 
> ```ruby
> # bad
> class Foo
>   def legal_name
>   end
>   
>   define_method :'illegal-name' do
>   end
> end
> 
> # good
> class Foo
>   define_method :legal_name do
>   end
>   
>   define_method :'illegal-name' do
>   end
> end
> ```
> 
> One additional argument against, is that I'm always cautious of rules where a later line could force changes in prior lines. So if the 5th line of a hash literal contains an illegal symbol literal name, it shouldn't cause git churn for lines 1-4 without good reason. I think the `=>`/`:` rises to that bar, but I don't think this does.
> 
> As a result, I'd be 👎 to making this change.

Fair points. To be honest, I don't have a good argument other than preferring consistency when reading a hash definition.

Due to configured `Style/HashSyntax` I can't even manually change
```ruby
{
  a: 1,
  "b-c": 2
}
```
to
```ruby
{
  :"a" => 1,
  :"b-c" => 2
}
```

🤷

aliismayilov:
I guess an example from a real project would have more weight to it :) See how it even messes up with syntax highlighting.

Screenshot 2024-01-12 at 19 32 52

camilopayan:
I don't think I understand how the Style/HashSyntax rule which controls the hashrocket usage and the Lint/SymbolConversion rule are causing an issue with each other. I _think_ they're covering different parts of the hash syntax.

I'll admit that the amount of code churn this would cause seems like enough to 👎

Issue #575, opened 2023-08-17
LSP: Autoformat on save in neovim


mna:
Hello,

I'm using neovim with the [standardrb LSP configured](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md#standardrb): 

```
$ nvim --version
NVIM v0.10.0-dev+852-g54be7d6b4
Build type: RelWithDebInfo
LuaJIT 2.1.0-beta3
```

It does report the `documentFormattingProvider` capability, but it doesn't seem to work when I try to format a buffer, I get the following error in the LSP log:

```
[ERROR][2023-08-17 10:54:09] .../vim/lsp/rpc.lua:675	"rpc"	"<path>/standardrb"	"stderr"	"[server] Format request arrived before text synchonized; skipping: `file:///<path>/a.rb'\n"
```

Is there some configuration I'm missing? I'm using the default LSP configuration as documented in the nvim-lspconfig link above (`require'lspconfig'.standardrb.setup{}`). I noticed the [stardard repo's Wiki on neovim](https://github.com/standardrb/standard/wiki/IDE:-neovim) doesn't mention formatting as part of its page, so I'm not sure if that's something that's supported or not (even though it does report the capability).

Thanks,
Martin

searls:
hey @will would you be so kind as to look at this?

mna:
I've taken a closer look at this, it's not just autoformat on save, it's the formatting of the buffer that has this bug (that is, it doesn't have to be called in a BufWritePre to reproduce).

If that helps (I'm not a ruby expert, nor an LSP one!), when I tweak that line https://github.com/standardrb/standard/blob/main/lib/standard/lsp/routes.rb#L163 for this:

```
@text_cache[file_uri] = text unless text.nil?
```

formatting does work. It seems like for some reason, it receives a `nil` text before formatting (even though the vim buffer is never empty) and that clears the `@text_cache` thing so no formatting happens.

fernandes:
@mna after testing this `unless`, right, it formats, but it does not show the diagnostics anymore

mna:
@fernandes oh no, for sure, I didn't mean that this was a workaround (and even less a fix)! Just that it illustrates that for some reason, the `diagnostic` call receives a `nil` text before formatting, storing it in the `@text_cache` and that's why it doesn't format anything.

will:
[...]

PR #601, opened 2024-01-30
Remove incorrect textDocument/diagnostic request handler

muxcmux:
The parameters for this request only include a `textDocument`, which is a `TextDocumentIdentifier`, an `identifier?` and a `previousResultId?`. The `TextDocumentIdentifer` only has a `uri` property, which is of type `DocumentUri`.

According to the LSP spec:

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_diagnostic

this request does not include the contents of the file (which is what the handler was expecting).

I have simply made the handling of this request a no-op, since diagnostics are already sent to the client in `textDocument/didChange`.

Note: This also fixes formatting problems related to Neovim 0.10 described in #575

### DIFF
--- a/lib/standard/lsp/routes.rb
+++ b/lib/standard/lsp/routes.rb
@@ -13,7 +13,7 @@ module Standard
       end
 
       def self.handle(name, &block)
-        define_method("handle_#{name}", &block)
+        define_method(:"handle_#{name}", &block)
       end
 
       def for(name)
@@ -51,10 +51,8 @@ module Standard
         end
       end
 
-      handle "textDocument/diagnostic" do |request|
-        doc = request[:params][:textDocument]
-        result = diagnostic(doc[:uri], doc[:text])
-        @writer.write(result)
+      handle "textDocument/diagnostic" do |_request|
+        # no op, diagnostics are handled in textDocument/didChange
       end
 
       handle "textDocument/didChange" do |request|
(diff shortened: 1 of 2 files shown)


searls:
Thanks for submitting this. Since VS Code in particular feels pretty black-box with respect to what it expects (and what it will reject) from an LSP server, I'd be reticent for this to be merged before someone first tested it against https://github.com/standardrb/vscode-standard-ruby

muxcmux:
I don't follow. This has nothing to do with vscode.

The handling of `textDocument/diagnostic` is not correct because the parameters do not include the document text - it's [literally in the specs](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentIdentifier).

searls:
Because I assumed (and [just confirmed](https://github.com/standardrb/standard/commit/1f75279aa878cba316b4263e143db8310e45007c)) this method was added because VS Code was spewing errors without it.

muxcmux:
I see. That method is still there though, I only made it a no-op. No errors in vscode as far as I can tell (fresh install of vscode, standardrb running with my patch):

https://github.com/standardrb/standard/assets/473227/a7826f3d-9afc-4b10-ba71-5418171d7a4c

... but I don't use vscode, so might not be a reliable test.

badboy666007 (approved):

PR #605, opened 2024-02-06
Disable Style/RedundantInitialize

### LINKED ISSUE #602: Removes empty initialize methods: Style/RedundantInitialize: Remove unnecessary empty `initialize` method.
Before:

```ruby
class A
  def initialize
    raise NotImplementedError
  end
end

class B < A
  def initialize
  end
end

puts B.new
```

After:

```ruby
# frozen_string_literal: true

class A
  def initialize
    raise NotImplementedError
  end
end

class B < A
end

puts B.new
```

```
> standardrb -v
1.33.0
```

czeise:
Fixes #602 

As pointed out in #602, this cop is unsafe and could change intended behavior.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -1566,7 +1566,7 @@ Style/RedundantHeredocDelimiterQuotes:
   Enabled: true
 
 Style/RedundantInitialize:
-  Enabled: true
+  Enabled: false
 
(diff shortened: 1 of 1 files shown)


camilopayan (approved):

dorianmariefr:
Thanks

Issue #603, opened 2024-02-02
Ruby 3.3 support


TastyPi:
Ruby 3.3 introduced an issue that causes `Naming/BlockForwarding` and `Style/ArgumentsForwarding` to break https://github.com/rubocop/rubocop/issues/12571. That issue has been fixed in RuboCop `1.60`, but this gem is pinned to `1.59`.

I believe all that's needed to support Ruby 3.3 is to set `spec.add_dependency "rubocop", "~> 1.60.0"`.

searls:
Looking now it looks like standard's current release is pegged to 1.61 so this should be resolved

PR #606, opened 2024-02-06
Upgrade rubocop

pythe:
I'm troubleshooting an issue with Layout/MultilineMethodCallIndentation. The first step is to use the latest Rubocop!

There were some formatting changes upon running `rake`, so I've left those in a separate commit.


pythe:
This may also address #603

pythe:
Handled in another PR. Thanks!

PR #614, opened 2024-03-12
bump rubocop to 1.62

robsdudeson:


searls:
Hey @robsdudeson, before cutting a release I think we should address https://github.com/standardrb/standard/issues/613

PR #619, opened 2024-03-21
Update Github Actions plugin in README.md

kmcphillips:
The originally recommended Github Action in the README has gone unmaintained and no longer works due to the Ruby version being too old. 

See:
* https://github.com/andrewmcodes/standardrb-action/pull/20
* https://github.com/andrewmcodes/standardrb-action/pull/23
* https://github.com/andrewmcodes/standardrb-action/commits/main/ (last commit 2 years ago)

I have forked it, updated the ruby version, tidied and cleaned up some out of date stuff, and added support for the [`standard-rails`](https://github.com/standardrb/standard-rails) plugin.

I've published it to the marketplace: https://github.com/marketplace/actions/standardrb-linter

I don't mind maintaining it going forward. There's not much code in there.

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,7 +184,7 @@ above command and will subsequently manage the server process for you.
 Various continuous integration and quality-checking tools have been made to
 support Standard Ruby, as well.
 
-* [Github Actions](https://github.com/andrewmcodes/standardrb-action)
+* [Github Actions](https://github.com/kmcphillips/standardrb-action) [(Marketplace)](https://github.com/marketplace/actions/standardrb-linter)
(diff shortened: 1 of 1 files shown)


searls:
Thanks for the heads up, but I believe this link should be to the official GH Action, no? https://github.com/standardrb/standard-ruby-action

cc/ @camilopayan @leequarella

kmcphillips:
Whelp! I did not find this official one. I followed the link in the README and tried to use and then fixed that one. If y'all as an org are going to maintain an action that's much better.

searls:
I think this is closable now? https://github.com/standardrb/standard-ruby-action

kmcphillips:
Yep! Sorry to leave it hanging.

PR #626, opened 2024-04-20
change regex percent literal delimiter to singlequote

Epigene:
Curly braces are a part of Regex syntax (`.{3,}`, `.{,9}` etc.),  so if the literal contains curlybraces, I have to use some other unspecified fallback delimiters. Bikeshedding ensues. Let's use delimiters that are not a part of Regex syntax and are less likely to appear in them - singlequotes. This goes nicely with string literals using doublequotes.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -1007,7 +1007,7 @@ Style/PercentLiteralDelimiters:
     default: ()
     '%i': '[]'
     '%I': '[]'
-    '%r': '{}'
+    '%r': "''"
     '%w': '[]'
     '%W': '[]'
 


searls:
Though I don't use it myself, my understanding is that `%r{}` is quite common and popular, so this would be a pretty major change. Additionally, I just scanned the regexes in the app that I'm building and a full third of them contain `'` characters, so this would be a pretty disruptive change AFAICT.

Would welcome other perspectives here

benjamineskola:
As I understand it, paired brackets as delimiters don't prevent you from using the same character, unescaped, within the literal:

```ruby
irb(main):001> %r{a{0,1}}
=> /a{0,1}/
```

Whereas unpaired characters like `'` would require that all `'` be escaped:

```ruby
irb(main):002> %r'it\'s'
=> /it's/
```

So since `%r{}` doesn't prevent any characters being used, and `%r''` would, I don't think this change would make sense; there's nothing to gain by it.

(In the worst-case scenario, a regex containing an unpaired closing `}` would require it to be escaped if using `%r{}`. But that would make it no worse than using `%r''` when the regex contains a `'`, and in any case `//`-literals are a viable alternative.)

searls:
Makes sense to me, failing a clear counter-argument. Thanks!

PR #630, opened 2024-05-17
Add support for Ruby LSP as a built-in add-on

searls:
## Release plan

- [x] Add a Ruby LSP addon as a formatter and a (pull) diagnostic provider
- [x] Add a test for the addon, using a modified test server helper that relies on ruby_lsp internals
- [x] Drop support for Ruby 2.7, so we can test the LSP plugin
- [x] Get CI passing again
- [x] Validate this all works in VS Code with the latest version of Ruby LSP
- [ ] Cut a patch release   

## Background

After talking to @vinistock at Kaigi, I want to take a stab at advertising a working Ruby LSP addon from within the gem so that Ruby LSP users can get standard support without running the LSP themselves (or installing a custom extension, like our [VS Code extension](https://marketplace.visualstudio.com/items?itemName=testdouble.vscode-standard-ruby)

I'm trying and failing to get this end-to-end working in an example app, so I'm trying to take a step back and just get this test passing. The result's response is nil, but (one assumes?) it shouldn't be.  More importantly, `WrapsBuiltinLspStandardizer#run_formatting` isn't being invoked, so my attempts to set the formatter to `standard is probably not working, either.

To replicate the failure, you should be able to run:

```
$ bundle
$ bundle exec m test/ruby_lsp_addon_test.rb
```

### DIFF
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -12,6 +12,7 @@ module RubyLsp
           @config,
           ::Standard::Lsp::Logger.new
         )
+        @rubocop_config = @config.rubocop_config_store.for_pwd
         @cop_registry = RuboCop::Cop::Registry.global.to_h
       end
 
@@ -73,7 +74,7 @@ module RubyLsp
       # https://github.com/Shopify/ruby-lsp/blob/4c1906172add4d5c39c35d3396aa29c768bfb898/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L84
       def code_description(cop_name)
         if (cop_class = @cop_registry[cop_name]&.first)
-          if (doc_url = cop_class.documentation_url)
+          if (doc_url = cop_class.documentation_url(@rubocop_config))
             Interface::CodeDescription.new(href: doc_url)
           end
         end
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -1,0 +1,89 @@
+require "lint_roller"
+require_relative "../../standard/builds_config"
+require_relative "../../standard/lsp/standardizer"
+require_relative "../../standard/lsp/logger"
+
+module RubyLsp
+  module Standard
+    class WrapsBuiltinLspStandardizer
+      include RubyLsp::Requests::Support::Formatter
+      def initialize
+        init!
+      end
+
+      def init!
+        @config = ::Standard::BuildsConfig.new.call(ARGV)
+        @standardizer = ::Standard::Lsp::Standardizer.new(
+          @config,
+          ::Standard::Lsp::Logger.new
+        )
+        @cop_registry = RuboCop::Cop::Registry.global.to_h
+      end
+
+      def run_formatting(uri, document)
+        # TODO: this isn't being triggered by my test
+        @standardizer.format(uri_to_path(uri), document.source)
+      end
+
+      def run_diagnostic(uri, document)
+        offenses = @standardizer.offenses(uri_to_path(uri), document.source)
+
+        offenses.map { |o|
+          cop_name = o[:cop_name]
+
+          msg = o[:message].delete_prefix(cop_name)
+          loc = o[:location]
+
+          severity = case o[:severity]
+          when "error", "fatal"
+            RubyLsp::Constant::DiagnosticSeverity::ERROR
+          when "warning"
+            RubyLsp::Constant::DiagnosticSeverity::WARNING
+          when "convention"
+            RubyLsp::Constant::DiagnosticSeverity::INFORMATION
+          when "refactor", "info"
+            RubyLsp::Constant::DiagnosticSeverity::HINT
+          else # the above cases fully cover what RuboCop sends at this time
+            logger.puts "Unknown severity: #{severity.inspect}"
+            RubyLsp::Constant::DiagnosticSeverity::HINT
+          end
+
+          RubyLsp::Interface::Diagnostic.new(
+            code: cop_name,
+            code_description: code_description(cop_name),
+            message: msg,
+            source: "Standard Ruby",
+            severity: severity,
+            range: RubyLsp::Interface::Range.new(
+              start: RubyLsp::Interface::Position.new(line: loc[:start_line] - 1, character: loc[:start_column] - 1),
+              end: RubyLsp::Interface::Position.new(line: loc[:last_line] - 1, character: loc[:last_column] - 1)
+            )
+            # TODO: do I need something like this?
+            # See: https://github.com/Shopify/ruby-lsp/blob/4c1906172add4d5c39c35d3396aa29c768bfb898/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L62
+            # data: {
+            #   correctable: @offense.correctable?,
+            #   code_actions: to_lsp_code_actions
+            # }
+          )
+        }
+      end
+
+      private
+
+      # duplicated from: lib/standard/lsp/routes.rb
+      def uri_to_path(uri)
+        uri.sub(%r{^file://}, "")
+      end
+
+      # lifted from:
+      # https://github.com/Shopify/ruby-lsp/blob/4c1906172add4d5c39c35d3396aa29c768bfb898/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L84
+      def code_description(cop_name)
+        if (cop_class = @cop_registry[cop_name]&.first)
+          if (doc_url = cop_class.documentation_url)
+            Interface::CodeDescription.new(href: doc_url)
+          end
+        end
+      end
+    end
+  end
+end
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -21,7 +21,6 @@ module RubyLsp
       end
 
       def run_formatting(uri, document)
-        # TODO: this isn't being triggered by my test
         @standardizer.format(uri_to_path(uri), document.source)
       end
 
@@ -72,7 +71,7 @@ module RubyLsp
 
       # duplicated from: lib/standard/lsp/routes.rb
       def uri_to_path(uri)
-        uri.sub(%r{^file://}, "")
+        uri.to_s.sub(%r{^file://}, "")
       end
 
       # lifted from:
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -53,12 +53,21 @@ module RubyLsp
               start: RubyLsp::Interface::Position.new(line: loc[:start_line] - 1, character: loc[:start_column] - 1),
               end: RubyLsp::Interface::Position.new(line: loc[:last_line] - 1, character: loc[:last_column] - 1)
             )
-            # TODO: do I need something like this?
+            # TODO: We need to do something like to support quickfixes thru code actions
             # See: https://github.com/Shopify/ruby-lsp/blob/4c1906172add4d5c39c35d3396aa29c768bfb898/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L62
             # data: {
-            #   correctable: @offense.correctable?,
+            #   correctable: correctable?(offense),
             #   code_actions: to_lsp_code_actions
             # }
+            #
+            # Right now, our offenses are all just JSON parsed from stdout shelling to RuboCop, so
+            # it seems we don't have the corrector available to us.
+            #
+            # Lifted from:
+            # https://github.com/Shopify/ruby-lsp/blob/8d4c17efce4e8ecc8e7c557ab2981db6b22c0b6d/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L201
+            # def correctable?(offense)
+            #   !offense.corrector.nil?
+            # end
           )
         }
       end
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -1,8 +1,3 @@
-require "lint_roller"
-require_relative "../../standard/builds_config"
-require_relative "../../standard/lsp/standardizer"
-require_relative "../../standard/lsp/logger"
-
 module RubyLsp
   module Standard
     class WrapsBuiltinLspStandardizer
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -7,7 +7,7 @@ module RubyLsp
       end
 
       def init!
-        @config = ::Standard::BuildsConfig.new.call(ARGV)
+        @config = ::Standard::BuildsConfig.new.call([])
         @standardizer = ::Standard::Lsp::Standardizer.new(
           @config,
           ::Standard::Lsp::Logger.new
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -51,7 +51,7 @@ module RubyLsp
             severity: severity,
             range: RubyLsp::Interface::Range.new(
               start: RubyLsp::Interface::Position.new(line: loc[:start_line] - 1, character: loc[:start_column] - 1),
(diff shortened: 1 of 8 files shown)


st0012:
It's failing because Ruby LSP doesn't format files [outside of the workspace path](https://github.com/Shopify/ruby-lsp/blob/2ce89a225e14bdea47280e5dfac579bb2d7cf039/lib/ruby_lsp/server.rb#L388). You can workaround that with:

```diff
-    with_server(source) do |server, uri|
+    with_server(source) do |server, _uri|
+      uri = URI(File.join(server.global_state.workspace_path, 'fake.rb'))
       server.process_message(
         id: 1,
-        method: "textDocument/formatting",
-        params: {textDocument: {uri: uri}, position: {line: 0, character: 0}}
+        method: 'textDocument/formatting',
+        params: { textDocument: { uri: uri },
+                  position: { line: 0, character: 0 } }
       )
```

However, it'd then actually try to run formatting with that file, which will result to another error since the file doesn't actually exist.

So you'll likely need additional setup to:
- Create an actual file with the target source (maybe under `path_to_standard/tmp/fake.rb`)?
- Run the test against that path
- Remove the file / or `tmp/` altogether

searls:
@st0012 thanks Stan! Got through that and have it formatting correctly now. 

Question: can I safely rely on `Dir.pwd` being the workspace directory, as I do with my custom extension?

Question: this test is now failing b/c the range being returned ends on line 19, column 19, even though the source string is only 2 lines long and that second line is only 5 or 6 characters. Since the addon is not providing the range, is it a bug?

searls:
I'm unsure the cause of the test failure this introduces, but just looking at the addon stuff, does this look right? @vinistock, @st0012?

searls:
Looking into this failure, I have very few good answers.

AFAICT, if one executes `require "ruby-lsp/internal"` in our test suite, then [the CLI test](https://github.com/standardrb/standard/blob/main/test/standard/cli_test.rb#L13) will start failing, with the odd behavior of suddenly being incapable of autofixing `Lint/TrailingCommaInAttributeDeclaration`

The test will print this to stderr:

```
tmp/cli_test/subject.rb:18:19: Lint/TrailingCommaInAttributeDeclaration: Avoid leaving a trailing comma in attribute declarations.
```

And then the test will fail because the shelled-out run will exit non-zero.

This is a bit of a mystery to me. Like, the best explanation I can imagine is that loading `ruby_lsp/internal` causes some Rubocop something-or-other to be loaded which in turn mutates the default configuration set of `Lint/TrailingCommaInAttributeDeclaration` so as to flag it as not autocorrectable?

Extremely odd failure

searls:
Hey @vinistock @st0012, I'm trying to validate this actually works by setting up a project pointing at my local copy using:

```ruby
  gem "standard", path: "/path/to/standardrb/standard"
```

But Ruby LSP doesn't seem to pick it up as a formatter/linter option.

screenshot-2024-05-23-11h16m16s

(If I `puts` something at the end of the addon's `activate` method, I'll see it blow up in the Ruby LSP output tab, so I know it's being required, however)

**Question:** do you have any idea why this might not be working when the plugin is a local path-sourced gem?

Meanwhile, if I change my dependency to look at this github branch:

```
gem "standard", github: "standardrb/standard", branch: "ruby-lsp-spike"
```

And then I check the output tab of VS Code, I get a stack trace without an error message:

```
2024-05-23 11:30:02.677 [info] (grog) /Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/standard/plugin/creates_runner_context.rb:7:in `call'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/standard/plugin/merges_plugins_into_rubocop_config.rb:29:in `call'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/standard/plugin/combines_plugin_configs.rb:11:in `call'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/standard/creates_config_store.rb:22:in `block in call'

<internal:kernel>:90:in `tap'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/standard/creates_config_store.rb:19:in `call'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/standard/builds_config.rb:32:in `call'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb:15:in `init!'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb:11:in `initialize'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/ruby_lsp/standard/addon.rb:15:in `new'

/Users/justin/.rbenv/versions/3.3.0/lib/ruby/gems/3.3.0/bundler/gems/standard-e7d5e49aafd2/lib/ruby_lsp/standard/addon.rb:15:in `activate'
```

So I'm kind of at a loss as to how to actually test the plugin locally, as a result.

st0012:
It looks like the error was caused by `Standard::VERSION` not being defined when it's called [here](https://github.com/standardrb/standard/blob/main/lib/standard/plugin/creates_runner_context.rb#L7).
Adding `require_relative "../../standard/version"` makes everything works as expected.

Screenshot 2024-05-23 at 22 31 16

And for some reason the addon's error reporting mechanism didn't work as expected as the error message is not displayed. I'll give it a look when I'm back to work.

andyw8:
> And for some reason the addon's error reporting mechanism didn't work as expected as the error message is not displayed.

This should be solved by https://github.com/Shopify/ruby-lsp/pull/2085

andyw8:
@searls You probably need to add `standard` to the enum [here](https://github.com/Shopify/ruby-lsp/blob/85b9e29d24c4eaed986a140aead82e8a436d3aad/vscode/package.json#L273).


But the extension shouldn't really have to know about specific addons, so we probably need to re-think this approach.

searls:
@andyw8 ahh, yes, if it's just hard-coded now that won't really scale to third-party add-ons. I was assuming the extension was phoning home to the LSP and then updating the configuration schema dynamically (somehow? Think I was conflating it with [WorkspaceConfiguration API](https://code.visualstudio.com/api/references/vscode-api#WorkspaceConfiguration), which I doubt can be used for that purpose.

It might be necessary to just relax that configuration property in the schema to allow it to be set to any value?

andyw8:
Looking into some options: https://github.com/Shopify/ruby-lsp/pull/2092

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
+module RubyLsp
+  module Standard
+    class WrapsBuiltinLspStandardizer
+      include RubyLsp::Requests::Support::Formatter
+      def initialize
+        init!
+      end
+
+      def init!
+        @config = ::Standard::BuildsConfig.new.call(ARGV)
```
vinistock, line 10: ARGV here is going to be the set of arguments passed to the `ruby-lsp` executable, which are likely not recognized by standard. Could that lead to any issues? Would it be better to pass an empty array every time?

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
...
+        @standardizer = ::Standard::Lsp::Standardizer.new(
+          @config,
+          ::Standard::Lsp::Logger.new
```
vinistock, line 13: Does this logger have any chance to print to stdout? In the LSP context, nothing can be printed to STDOUT or else is breaks editor/server communication. Everything has to be printed to STDERR only.

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
...
+        )
+        @cop_registry = RuboCop::Cop::Registry.global.to_h
+      end
+
+      def run_formatting(uri, document)
+        @standardizer.format(uri_to_path(uri), document.source)
```
vinistock, line 19: We have a helper that patches a new method into URI objects to get the correct file system paths, which has some fixes for Windows (the URI gem has some bugs).

```suggestion
        @standardizer.format(uri.to_standardized_path, document.source)
```

There's only one caveat. If you created a Ruby file and it is currently unsaved, you can invoke formatting manually. In those cases, `to_standardized_path` returns `nil`. I'm not sure how Standard will handle that.

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
...
+      end
+
+      def run_diagnostic(uri, document)
+        offenses = @standardizer.offenses(uri_to_path(uri), document.source)
```
vinistock, line 23:
```suggestion
        offenses = @standardizer.offenses(uri.to_standardized_path, document.source)
```

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
...
+          when "error", "fatal"
+            RubyLsp::Constant::DiagnosticSeverity::ERROR
+          when "warning"
+            RubyLsp::Constant::DiagnosticSeverity::WARNING
+          when "convention"
+            RubyLsp::Constant::DiagnosticSeverity::INFORMATION
+          when "refactor", "info"
+            RubyLsp::Constant::DiagnosticSeverity::HINT
+          else # the above cases fully cover what RuboCop sends at this time
+            logger.puts "Unknown severity: #{severity.inspect}"
```
vinistock, line 41: Does this print to STDOUT? If so, it will throw the LSP into an infinite loop.

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
...
+            code: cop_name,
+            code_description: code_description(cop_name),
+            message: msg,
+            source: "Standard Ruby",
+            severity: severity,
+            range: RubyLsp::Interface::Range.new(
+              start: RubyLsp::Interface::Position.new(line: loc[:start_line] - 1, character: loc[:start_column] - 1),
+              end: RubyLsp::Interface::Position.new(line: loc[:last_line] - 1, character: loc[:last_column] - 1)
+            )
+            # TODO: do I need something like this?
```
vinistock, line 55: If you want to support quickfixes through code actions, then yes you'll need to provide those as part of data.

But you can always start without it and add later.

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -0,0 +1,83 @@
...
+      private
+
+      # duplicated from: lib/standard/lsp/routes.rb
+      def uri_to_path(uri)
+        uri.to_s.sub(%r{^file://}, "")
+      end
+
+      # lifted from:
+      # https://github.com/Shopify/ruby-lsp/blob/4c1906172add4d5c39c35d3396aa29c768bfb898/lib/ruby_lsp/requests/support/rubocop_diagnostic.rb#L84
+      def code_description(cop_name)
```
vinistock, line 74: I should warn you that a contributor is looking into changing this slightly to allow for RuboCop extensions to also have their documentation URLs displayed on diagnostics.

Likely not something that needs to be addressed here, just an FYI https://github.com/Shopify/ruby-lsp/pull/2076.

vinistock (commented):
❤️

andyw8:
@searls `standard` is now available as a formatter option: https://github.com/Shopify/ruby-lsp/pull/2092

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
searls: The [Logger](/lib/standard/lsp/logger.rb) is careful to only output over stderr, to keep the stdout communication channel clean

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
searls: The [Logger](/lib/standard/lsp/logger.rb) is careful to only output over stderr, to keep the stdout communication channel clean

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
searls: I suppose I'll just wrap my own helper and guard against the nil case to fall back:

```ruby
      def uri_to_path(uri)
        if uri.respond_to?(:to_standardized_path) && !(standardized_path = uri.to_standardized_path).nil?
          standardized_path
        else
          uri.to_s.sub(%r{^file://}, "")
        end
      end
```

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
searls: Thanks for the heads up, we already depend on rubocop ~> 1.64.0 so I can just resolve the RuboCop config store `for_pwd` and wedge that argument in, it seems

lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
searls: Thanks for explaining. Right now I only have access to a JSON representation of the offense parsed from RuboCop over stdout, so getting the corrector will (I assume?) require us to invoke RuboCop directly in our LSP code.

searls:
@andyw8 is there any way I ought to be able to test this now?

I tried pointing `"rubyLsp.branch": "main"` since the gem hasn't been released since you merged https://github.com/Shopify/ruby-lsp/pull/2092, but of course that's based on the vscode extension's package.json, so then I cloned and ran _that_ in a host extension but, try as I might, even though I can get the following config to clear warnings/errors:

```json
{
  "[ruby]": {
    "editor.defaultFormatter": "Shopify.ruby-lsp"
  },
  "rubyLsp.formatter": "standard",
  "rubyLsp.linters": [
    "standard"
  ]
}
```

… I don't get any diagnostics and I still get `"There is no formatter for 'ruby' files installed."` when I try to manually invoke the Format Document command

searls:
Just released this basic version (without code actions) as v1.38.0

searls:
@vinistock @andyw8 if you would be so kind, let me know when a version of the vscode extension has shipped with the `standard` enum present and I'll go to town testing this more thoroughly, adding it to our docs, and so forth

andyw8:
@searls you can test it by checking out the `main` branch, selecting the `vscode` workspace and, then selecting `Run and Debug`:

https://github.com/Shopify/ruby-lsp/blob/main/CONTRIBUTING.md#live-debugging

Everyone on the team is at a company event this week, so there may be a little delay until the next release.

searls:
Heads up I also merged #636 which adds Code Action support (and duplicates half of Ruby LSP's rubocop integration in the process)

andyw8:
>@vinistock @andyw8 if you would be so kind, let me know when a version of the vscode extension has shipped with the standard enum present and I'll go to town testing this more thoroughly, adding it to our docs, and so forth

This shipped today: https://github.com/Shopify/ruby-lsp/releases/tag/vscode-ruby-lsp-v0.7.6

searls:
Thanks @andyw8. I've verified everything works, and so have updated these docs:

* [Re-introduced a VS Code wiki page](https://github.com/standardrb/standard/wiki/IDE:-vscode#using-ruby-lsp) on Standard's repo documenting setup
* Highlight the Ruby LSP add-on [on the README of our VS Code extension](https://github.com/standardrb/vscode-standard-ruby), pointing out that it additionally supports code actions
* [Update the Standard README](https://github.com/standardrb/standard?tab=readme-ov-file#language-server-protocol-support) explaining both options, steering people to Ruby LSP for its pull diagnostics and code actions support 

I also [just published a blog post](https://justin.searls.co/posts/why-i-just-uninstalled-my-own-vs-code-extension/) that I hope you all will appreciate as supporting the vision that @vinistock and others have laid out for the Ruby LSP project.

Additionally, I'd love if you highlights this addon on your repo's README [along with the others](https://github.com/Shopify/ruby-lsp?tab=readme-ov-file#addons), since the fact that this one is built-in will make discovery harder for Ruby LSP users (especially in the presence of our own bespoke LSP server and VS Code plugin).

st0012:
I've created https://github.com/Shopify/ruby-lsp/pull/2263 to include standard in the readme 👍

vinistock:
@searls thank you very much for the blog post and your work on this. Having your support for the vision is really important and I appreciate it 🙏.

PR #634, opened 2024-06-16
Restore 3.0 to CI, bump gemspec

mjankowski:
Context: https://github.com/standardrb/standard-performance/pull/26#issuecomment-2171548753

Prep for possible aligned release of standard/-rails/-performance

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,6 +17,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         ruby-version:
+          - '3.0'
           - '3.1'
           - '3.2'
           - '3.3'
--- a/config/default.yml
+++ b/config/default.yml
@@ -1,6 +1,6 @@
 AllCops:
-  # Prevent RuboCop from exploding when it finds an older-than-2.6 .ruby-version
-  TargetRubyVersion: 2.6
+  # Prevent RuboCop from exploding when it finds an older .ruby-version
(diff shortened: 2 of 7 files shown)


mjankowski:
Added commit to also bump `TargetRubyVersion` (RE: https://github.com/standardrb/standard-performance/pull/26#issuecomment-2171671971)

mjankowski:
Another possible update here ... looks like the `Standard::Base::Plugin` class has a method that does yaml merging for backwards support of many ruby versions, with each version inheriting from the next highest one and disabling the cops added in that higher version. So probably...

- If we are doing a hard drop of support for pre-3.0 ... can we remove all the 1.x and 2.x yaml files and conditional cases in that plugin file?
- We probably need a 3.3 yml added, and to update 3.2 yml (if any relevant drops exist, I'm not sure yet)

searls:
@mjankowski no, those are all there intentionally. Standard is designed to allow users to target ancient Rubies. If you run Standard under a supported-by-rubocop Ruby against a 1.8.7 codebase by setting TargetRubyVersion, it'll still be able to lint/format without breaking 1.8.7 compatibility

mjankowski:
Oh right - disconnect between runtime support and linting target support. Got it.

Pushed another commit here which adds 3.3 yml and adjusts 3.2 in what I think is correct mirroring other files. As far as I can see there are not any cops which need 3.3 as minimum.

PR #635, opened 2024-06-23
Fix end position of diagnostic for LSP

koic:
Follow up https://github.com/rubocop/rubocop/pull/12932.

This is a porting because it can be said that even Standard.rb, which serves as the base for RuboCop's built-in LSP, does the same thing.

---

This is a quote from https://github.com/rubocop/rubocop/pull/12932.

`character` of `Position` is zero-based
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
> Character offset on a line in a document (zero-based).

And, `Parser::Source::Range` is also zero-based
https://github.com/whitequark/parser/blob/3e260d2e37bcb3de8705489d1c2799c26c7a2215/lib/parser/source/range.rb#L104
> zero-based column number of the end of this range.

Therefore, `-1` is not necessary.


searls:
Thanks for back/forward/down porting this for us -- I would've missed it otherwise. Incidentally, I was looking at the exact same line of code in my spike of a Ruby LSP addon today, so I'll fix this there too.

PR #639, opened 2024-07-02
Update ci workflow versions

mjankowski:
Should resolve the two node version warnings in output of update task - https://github.com/standardrb/standard/actions/runs/9736076397

There are other deprecations there re: the `set-output` command, and a failure in latest run. May check those out separately.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
       - name: Set up Ruby
         uses: ruby/setup-ruby@v1
         with:
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -45,7 +45,7 @@ jobs:
         id: date
         run: echo "::set-output name=date::$(date +'%Y-%m-%d')"
       - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


mjankowski:
The failure on the workflow run might be perms related? - https://github.com/standardrb/standard/actions/runs/9736076397/job/26866177750#step:10:217

```
remote: Permission to standardrb/standard.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/standardrb/standard/': The requested URL returned error: 403
Error: The process '/usr/bin/git' failed with exit code 128
```

searls:
Maybe this is an older way of doing stuff? I haven't looked at that workflow before, but recently I've been seeing explicit repo-level permissions like this (in [Standard's own official action](https://github.com/standardrb/standard-ruby-action?tab=readme-ov-file#creating-a-standard-ruby-workflow)):

```yml
jobs:
  build:
    runs-on: ubuntu-latest
    permissions:
      checks: write
      contents: write
    steps:
    - name: Standard Ruby
      uses: standardrb/standard-ruby-action@v1
```

mjankowski:
I have also not looked at that workflow in detail and am half-guessing here.

The logs and records only go back so far - https://github.com/standardrb/standard/actions/workflows/update.yml - but its been failing for a while.

Does the `github-actions` user have the correct perms at the org/repo level? https://github.com/peter-evans/create-pull-request?tab=readme-ov-file#workflow-permissions

PR #641, opened 2024-07-04
Add Datadog to the list of teams using standard

anmarchenko:
We love standardrb at Datadog, so I would like to add ourselves to the list of teams adopting your styleguide.

Thank you for the great work!


searls:
Woot! Love to hear it, thanks @anmarchenko! 🐶💎

PR #642, opened 2024-07-16
Ensure returned uri from code actions contain a scheme

jimnanney:
Using Neovim 0.10 with the ruby-lsp plugin, standard returns code actions to allow for quick fixes. They were returning the uri without the scheme, and that caused neovim to give an error.

To correct for the error, I added an ensure_uri_scheme method to parse the uri and add the "file" scheme if the uri did not include one.

Example of the erorr:
image

Edit: Updated tests to use new file type uri in return.

### DIFF
--- a/lib/standard/lsp/diagnostic.rb
+++ b/lib/standard/lsp/diagnostic.rb
@@ -82,7 +82,7 @@ module Standard
             document_changes: [
               Interface::TextDocumentEdit.new(
                 text_document: Interface::OptionalVersionedTextDocumentIdentifier.new(
-                  uri: @uri.to_s,
+                  uri: ensure_uri_scheme(@uri.to_s).to_s,
                   version: nil
                 ),
                 edits: correctable? ? offense_replacements : []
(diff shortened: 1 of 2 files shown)


searls:
Hi Jim!

Have you pointed your neovim at this to verify it works manually?

jimnanney:
yes, I updated in my neovim copy of standardrb to test it manually. I don't have vscode, so I was a little afraid of maybe breaking vscode.

searls:
I'd be extremely shocked if this broke anything so happy to merge

searls:
Released as 1.39.2

If anyone has a chance to test quickfixes (command-period) on VS Code to see if it fixes it there, we should update the table in the README https://github.com/standardrb/standard?tab=readme-ov-file#language-server-protocol-support

jimnanney:
Wow, thanks for merging and releasing that so quickly! Really appreciate it!

searls:
You break it, you buy it!

PR #643, opened 2024-07-17
Update Ruby LSP version

andyw8:
Just to prevent this warning when working on standard itself:

Screenshot 2024-07-17 at 8 40 47 AM


andyw8:
Fixed. We'll try find a better way so that those all don't need to manually required in your `ruby_lsp_addon_test.rb`.

PR #651, opened 2024-09-26
Add Privy to Standard users list

spencern:
Privy has adopted Standard!


searls:
So happy to have you in the club, Spencer!

PR #655, opened 2024-10-16
Add range formatting method to the formatter add-on

vinistock:
We added support for range formatting in the Ruby LSP, which allows users to format a partial selection of the document (it also powers format on paste).

Currently, we only added complete support for Syntax Tree, since it only formats and doesn't have linting rules. For the time being, we can avoid [this error](https://github.com/Shopify/ruby-lsp/issues/2726) by just defining the method and returning `nil`.

It may be possible to support range formatting with Standard, but I'm not sure exactly how. We'd need to understand which corrections apply to the selection and all style/layout corrections have to start with the `base_indentation` instead of considering zero.

### DIFF
--- a/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
+++ b/lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
@@ -20,6 +20,13 @@ module RubyLsp
         @standardizer.offenses(uri_to_path(uri), document.source, document.encoding)
       end
 
+      def run_range_formatting(_uri, _partial_source, _base_indentation)
+        # Not yet supported. Should return the formatted version of `partial_source` which is a partial selection of the
+        # entire document. For example, it should not try to add a frozen_string_literal magic comment and all style
+        # corrections should start from the `base_indentation`
+        nil
+      end
+
       private
 
       # duplicated from: lib/standard/lsp/routes.rb


lib/ruby_lsp/standard/wraps_built_in_lsp_standardizer.rb
```diff
@@ -20,6 +20,13 @@ def run_diagnostic(uri, document)
         @standardizer.offenses(uri_to_path(uri), document.source, document.encoding)
       end
 
+      def run_range_formatting(_uri, _partial_source, _base_indentation)
+        # Not yet supported. Should return the formatted version of `partial_source` is a partial selection of the
```
andyw8, line 24: Missing a word here?
```suggestion
        # Not yet supported. Should return the formatted version of `partial_source` which is a partial selection of the
```

searls:
A couple questions:

* Are you adding this to avoid a runtime error in Ruby LSP? Does this need to be merged urgently as a result?
* To be clear, Ruby LSP's RuboCop formatter doesn't support this feature, right? Just SourceTree?
* @koic, just to double-check, is RuboCop capable of range formatting yet via API? I get the sense that the way the cop API works, this would be very very difficult to implement without breaking compatibility or else require a plugin like this one to try to trim start and end points intelligently via a heuristic or something.

vinistock:
It breaks in the runtime, but it's not critical. There are only two ways to trigger the range formatting request: by manually telling the editor to format a range or by turning on format on paste. So it's not critical, but it does error in the runtime. Also, there's no detriment to returning `nil`, it will simply not format the range.

Only Syntax Tree supports this. It's more difficult for RuboCop (and I assume Standard too) because you will need a way to identify which rules only apply when analyzing the entire file. Since Syntax Tree has an API for "please format this node for me", without any assumptions that it's the complete file, all you need to adjust is the base indentation so that it knows you're starting from a deeper level.

RuboCop and Standard have linting rules in addition to formatting, so this is where I expect challenges from. There are rules that don't apply to a partial selection and could potentially produce incorrect corrections. For example, we have the auto correction for the frozen string literal comment enabled. If we simply ask RuboCop to format a selection, it will add the comment to that selection because it thinks it's formatting an entire file. E.g.:

```ruby
# BEFORE
# frozen_string_literal: true

a    + 5
# Select only the above line to format

# After
# frozen_string_literal: true

# frozen_string_literal: true
a + 5
```

We would need an API to apply auto corrections excluding rules that only make sense for the entire file and that accepts a base indentation as the start level for layout rules.

searls:
Makes sense, and I don't mind adding this. In the future, the add-on ecosystem might be well-served by having the runtime only attempt to invoke actions that the addon passes a `respond_to?` check, so that a routine extension update doesn't break Standard etc

PR #656, opened 2024-10-29
Proposal: Update rubocop and handles a new cop

tk0miya:
Since v1.67.0, the `AllowRBSInlineAnnotation` config option has been added to the `Layout/LeadingCommentSpace` cop.
It is necessary to annotate types in Ruby code via [RBS::Inline](https://github.com/soutaro/rbs-inline). RBS::Inline is a pilot project of [ruby/rbs](https://github.com/ruby/rbs) and is planned to be merged into ruby/rbs.

Could you update rubocop to v1.67.0 and enable the `AllowRBSInlineAnnotation` option, please?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+* Updates rubocop to [1.67.0](https://github.com/rubocop/rubocop/releases/tag/v1.67.0)
+
 ## 1.41.1
 
 * Adds a stub method to the Ruby LSP add-on to avoid a potential runtime exception
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,7 +4,7 @@ PATH
     standard (1.41.1)
       language_server-protocol (~> 3.17.0.2)
       lint_roller (~> 1.0)
-      rubocop (~> 1.66.0)
+      rubocop (~> 1.67.0)
       standard-custom (~> 1.0.0)
       standard-performance (~> 1.5)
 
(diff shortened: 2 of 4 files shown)


nicowenterodt:
I also see problems with the current 1.66 Rubocop version.

This is because of the latest update of [psych](https://github.com/ruby/psych/pull/686).

Resulting in errors like `An error occurred while Layout/FirstArgumentIndentation cop was inspecting` 

https://github.com/rubocop/rubocop/pull/11060#issuecomment-2467457040

This should be fixed when upgrading to 1.68.0 as this PR has made in to that version already: https://github.com/rubocop/rubocop/pull/13379

kpumuk:
Another option is `AllowSteepAnnotation`, which also needs to be enabled. With Steep 1.9, empty arrays will trigger a warning like:

```
lib/meta_tags/text_normalizer.rb:195:102: [warning] Empty array doesn't have type annotation
│ Diagnostic ID: Ruby::UnannotatedEmptyCollection
│
└         title = (title_limited_length > 0) ? truncate_array(title, title_limited_length, separator) : []
```

Standard will complain with

```
  lib/meta_tags/text_normalizer.rb:162:19: Layout/LeadingCommentSpace: Missing space after `#`.
```

Looks like current option is to either ditch (somewhat) Standard, or Steep. https://github.com/kpumuk/meta-tags/pull/323

jasonkarns:
fixed by https://github.com/standardrb/standard/pull/709

PR #657, opened 2024-11-03
Add doc around ruby version support/maintenance

mjankowski:
Background: https://github.com/standardrb/standard-custom/pull/7#issuecomment-2453470815


searls:
Good start. @camilopayan could you take a look?

Two requests: 

* Branding: please refer to the library as "Standard Ruby" or "Standard" (when Ruby is clear from the sentence of the context)
* Rather than list in a sentence, could we do a bullet point list of every relevant repo under the org? https://github.com/standardrb

Thanks!

mjankowski:
Pushed two commits to a) use correct naming/titling, b) add list of gems.

Just to confirm on full desired repo list -- I assume you want the four in there right now, and then ALSO `lint_roller` and `-sorbet`? Those latter two probably need some repo maintenance, can do that as followup once confirmed.

searls:
Yes we should do our best to keep them all in sync. I realize more repos -> more problems and that's 100% my fault for splitting standard up

mjankowski:
Added those to this list, will do PRs for them to bump up to 3.0 min in their own repos.

searls:
Yep. Let's definitely not move lint_roller up. I'm of two minds on whether to keep our compatibility with Rubocop's in lock-step as policy (even if we keep it 3 & up from now) but I don't feel as strongly

mjankowski:
Removed lint-roller from the list in this PR.

searls:
One idea: should this document have in bold at the top: "the current minimum supported ruby version is 3.0"? so when we link future PRs/issues about version bumps we can point to this without getting into a meta discussion of how we're arriving to the answer?

mjankowski:
Maybe? Might be a nice context shortcut to link back from future PRs on the various repos ... but on the other hand it's one more thing to keep current and it being linked to relies on anyone knowing it exists. Don't feel strongly either way really.

In either case, can definitely link back to this doc from ~6 PRs in various repos in a few months when presumably ruby 3.4.0 has been released, works fine, can be supported, and 3.0 (will be in the "past EOL but still supported" stage) could be dropped with a release to ring in the new year.

searls:
Gonna call this good enough without other feedback

mjankowski:
Cool - no prob, thanks for merge.

I think things are up to date now ... but will sync all my forks and do one more pass across the repos to check.

PR #660, opened 2024-11-12
Disable Style/ArgumentsForwarding

### LINKED ISSUE #612: Style/ArgumentsForwarding seems a bit heavy-handed
Just updated standard to 1.34.0 and saw this failure:

```ruby
  def set_time_zone(&block)
    Time.use_zone(current_user.time_zone, &block)
  end
```

Which auto-fixes to this:

```ruby
  def set_time_zone(&)
    Time.use_zone(current_user.time_zone, &)
  end
```

I was kinda annoyed by this, and maybe even more annoyed when I realized that only `[blk, block, proc]` [trip the rule](https://docs.rubocop.org/rubocop/cops_style.html#styleargumentsforwarding).

Am I just being an old crank because I have 20 years of muscle memory writing `&blk`?

bjeanes:
Closes #612

See https://github.com/standardrb/standard/issues/612#issuecomment-2467897059.

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -984,7 +984,7 @@ Style/AndOr:
   Enabled: true
 
 Style/ArgumentsForwarding:
-  Enabled: true
+  Enabled: false
 
 Style/ArrayCoercion:
   Enabled: false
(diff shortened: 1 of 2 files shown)


searls:
Thanks!

PR #668, opened 2024-11-19
Add test for loading config without warnings

mjankowski:
Background: https://github.com/standardrb/standard/pull/663#issuecomment-2480751535

This works in the sense that if I re-create that typo this test fails, and if I leave the typo fixed this passes, so it presumably would have caught that issue.

I'm not super-familiar with rubocop internals, and am very open to feedback here on more elegǎnt way to capture this. I was expecting to be able to either have a validation object with an activemodel-style `errors` array to assert against, and/or be able to just let this run and have the `ValidationError` be raised - but I think that's swallowed either somewhere within rubocop or somewhere in test harness.

### DIFF
--- a/test/standard/base_test.rb
+++ b/test/standard/base_test.rb
@@ -34,6 +34,13 @@ class Standard::BaseTest < UnitTest
     assert_equal expected, actual, "Cop names should be alphabetized! (See this script to do it for you: https://github.com/standardrb/standard/pull/222#issue-744335213 )"
   end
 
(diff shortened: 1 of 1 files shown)


mjankowski:
Yeah - I re-created the previous EMABLED typo, messed around with various rubocop config objects until something failed (looked for validation/errors objects at first, settled on this), then confirmed it passed when corrected.

searls:
Nice! Then it certainly won't hurt. And if it somehow does we can always delete it

PR #684, opened 2024-11-23
Add Nova to list of supported editors

edwardloveall:
This adds a link to the Nova extension for standard.


searls:
Awesome, thanks!

PR #690, opened 2025-01-14
add Lobsters to list of projects

pushcx:
I thought it'd be more interesting to link to our GitHub repo than https://lobste.rs, but I can swap if you prefer that.


searls:
Fun!

PR #706, opened 2025-03-25
Allow adjacent oneline method definitions

### LINKED ISSUE #695: Add AllowAdjacentOneLineDefs: true to Layout/EmptyLineBetweenDefs
I propose a rule change which would allow visual grouping of oneline methods, which can convey semantic meaning.

The relevant setting `AllowAdjacentOneLineDefs` is under [Layout/EmptyLineBetweenDefs](https://docs.rubocop.org/rubocop/1.72/cops_layout.html#layoutemptylinebetweendefs)

[This setting](https://github.com/standardrb/standard/blob/ee40a679e4ab21f6d8b2eb65f8e739d2b939b430/config/base.yml#L135) appears to have been [part of a large import of default Rubocop settings](https://github.com/standardrb/standard/commit/ff58f60a871c952b9e2971843b6c78de3c4cb705) and not an explicitly decided and a settled matter. It also sounds like this was at one time a rubocop default, and that does not appear to be the case today when I ran standardrb and rubocop autofixing over the same code. It may be that the default changed upstream.

A contrived example of what happens today:

Given this code:
```ruby
def isbn = lookup(“isbn”)
def barcode = lookup(“barcode”)
def sku = Skus.find(isbn)

def title = lookup(“title”) || “unknown”
def author = lookup(“author”)
```

standardrb would format it today as:

```ruby
def isbn = lookup(“isbn”)

def barcode = lookup(“barcode”)

def sku = Skus.find(isbn)

def title = lookup(“title”) || “unknown”

def author = lookup(“author”)
```

Not only does this double spacing make the code longer - and you'd expect single-line methods to be used for terseness - but the method grouping is lost, which conveys less information to the developer. My example is contrived, but I have seen this formatting make real, production code harder to read also.

The proposed rule change would have kept the original code the same.

soulcutter:
The purpose of oneline methods is a terser format for simple methods, similar to defining variables.

This rule change allows oneline methods to be grouped more-closely together, just as you might group variables to convey meaning. This also accomplishes terser class definitions that don't have mandatory double-spacing between oneline methods.

This permits code like:
```ruby
def one = "one"
def two = "two"

def number? = true
def animal? = false
```

Fixes #695

### DIFF
--- a/config/base.yml
+++ b/config/base.yml
@@ -132,7 +132,7 @@ Layout/EmptyLineAfterMultilineCondition:
 
 Layout/EmptyLineBetweenDefs:
   Enabled: true
-  AllowAdjacentOneLineDefs: false
+  AllowAdjacentOneLineDefs: true
   NumberOfEmptyLines: 1
 
 Layout/EmptyLines:


kmcphillips (commented):
+1

PR #726, opened 2025-06-12
Rubocop YAML schema

joeldrapper:
I’m writing a comprehensive schema for Rubocop in a new project called [Ruby Schema](https://github.com/joeldrapper/rubyschema). The goal of this project is to provide schemas for popular Ruby gems that use YAML configuration files.

The [YAML Language Server](https://github.com/redhat-developer/yaml-language-server) uses these schemas to provide auto-complete, hover documentation and validation. It’s available in loads of text editors and it’s installed by default in [Zed](https://zed.dev). Here’s an example from standard.

Screenshot 2025-06-12 at 11 02 04

The Rubocop schema is not finished yet — I’m writing this by hand! — but it is designed to allow for unknown cops so it’s already compatible with Standard’s configs.

It currently provides auto-complete for:

1. top-level configuration,
2. AllCops,
3. departments,
4. generic (shared) cop configuration, and
5. specific cop configuration for Bundler, Gemspec and up to the migration cops (still have naming, security and style to complete).

It’s still a very new project, I only started it yesterday so I understand any reluctance to merge this, but I wanted to at least raise awareness and invite feedback from the standard community.

❤️

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,3 @@
+{
+  "singleQuote": true
+}
--- a/.standard.yml
+++ b/.standard.yml
@@ -1,3 +1,7 @@
+# yaml-language-server: $schema=https://www.rubyschema.org/standard.json
+
+%YAML 1.1
+---
 ruby_version: 3.0
 ignore:
   - tmp/**/*
--- a/config/base.yml
+++ b/config/base.yml
@@ -1,3 +1,7 @@
+# yaml-language-server: $schema=../../rubyschema/dist/rubocop.json
+
+%YAML 1.1
+---
 Bundler/DuplicatedGem:
   Enabled: true
   Include:
--- a/config/base.yml
+++ b/config/base.yml
@@ -1231,7 +1231,7 @@ Style/EvenOdd:
   Enabled: false
 
 Style/ExactRegexpMatch:
-    Enabled: true
+  Enabled: true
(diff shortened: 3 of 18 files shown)


camilopayan:
I get the appeal! But our YAML files aren't really meant to be edited by anyone but standard maintainers, so I'm not sure this would be all that useful in our project. Seems like it would be killer for those rolling their own rubocop configurations though.

jasonkarns:
I have a couple questions on the how the schemas are referenced.

I assume from the full absolute URL that one could change the url (effectively at any time) and the YML language server would work with the schema pulled from anywhere? So if standard made changes that didn't conform to the existing schema, we could change the url and point to a json schema somewhere else that fully reflected those changes? (Like pointing to a file under our control for instance)

And secondly, does the schema reference support versioning of any kind? If the json schemas evolve, how are those references pinned to older (or future) versions?

joeldrapper:
Yes, the YAML language server will pull the schema from any URL. There is no versioning yet, because this is the first release. I plan to maintain this URL as the latest schema and will have new urls for old schemas when there are breaking changes.

The project is licensed MIT no-attribution, so you could just copy/paste the current schema at any time to customise it. But I would encourage you to contribute those changes back to the RubySchema project instead so other people can benefit from them.

We also maintain a schema for Standard’s own configuration here. https://www.rubyschema.org/standard.json

PR #730, opened 2025-07-15
Harden and enhance test workflow

jasonkarns:
- **Enable dependabot for all gem deps**
- **Enable dependabot for github-actions**
- **Run tests on schedule and on demand**

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -2,33 +2,23 @@ name: Tests
 
 on:
   push:
-    branches: [main]
   pull_request:
-    branches: [main]
+  workflow_dispatch:
+  schedule: [ cron: "42 0 7 * *" ] # https://crontab.guru/#42_0_7_*_*
   workflow_run:
-    workflows: ["Update"]
-    types:
-      - completed
+    workflows: [ "Update" ]
+    types: [ completed ]
 
 jobs:
   test:
-    strategy:
-      matrix:
-        os: [ubuntu-latest]
-        ruby-version:
-          - "3.1"
-          - "3.2"
-          - "3.3"
-          - "3.4"
-
-    runs-on: ${{ matrix.os }}
-
+    runs-on: ubuntu-latest
+    strategy: {matrix: {ruby: [ '3.0', 3.1, 3.2, 3.3, 3.4 ]}}
     steps:
-      - uses: actions/checkout@v4
-      - name: Set up Ruby ${{ matrix.ruby-version }}
-        uses: ruby/setup-ruby@v1
+      - uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with: { egress-policy: audit }
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      - uses: ruby/setup-ruby@a4effe49ee8ee5b8b5091268c473a4628afb5651 # v1.245.0
         with:
-          ruby-version: ${{ matrix.ruby-version }}
+          ruby-version: ${{ matrix.ruby }}
           bundler-cache: true
-      - name: Run tests for Ruby ${{ matrix.ruby-version }} on ${{ matrix.os }}
-        run: bundle config unset deployment && ./bin/rake
+      - run: bin/rake


.github/dependabot.yml
```diff
@@ -1,10 +1,11 @@
...
+# https://docs.github.com/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file
+
 version: 2
 updates:
-  - package-ecosystem: "bundler"
+  - package-ecosystem: bundler
     directory: "/"
-    allow:
-      - dependency-name: "rubocop"
-      - dependency-name: "rubocop-performance"
```
jasonkarns, line 7: Removed limitation to rubocop deps. This should allow other deps (mostly transitive deps) to get bumps in our lockfile (which only affects the maintainers, but keeps our lockfile up to date with latest).

.github/dependabot.yml
```diff
@@ -1,10 +1,11 @@
...
-    schedule:
-      interval: "weekly"
-      day: monday
+    schedule: { interval: weekly, day: monday }
```
jasonkarns, line 8: Monday is chosen so that deps are bumped approx 1 day before our regular standard update cadence.

.github/workflows/test.yml
```diff
@@ -2,33 +2,23 @@ name: Tests
 
 on:
   push:
-    branches: [main]
   pull_request:
-    branches: [main]
+  workflow_dispatch:
```
jasonkarns, line 6: This allows manually triggering test runs from the github UI (or github CLI). Not a frequent occurrence, but very handy to already have ready when needed.

.github/workflows/test.yml
```diff
@@ -2,33 +2,23 @@ name: Tests
...
-
+    runs-on: ubuntu-latest
+    strategy: {matrix: {ruby: [ '3.0', 3.1, 3.2, 3.3, 3.4 ]}}
     steps:
-      - uses: actions/checkout@v4
-      - name: Set up Ruby ${{ matrix.ruby-version }}
-        uses: ruby/setup-ruby@v1
+      - uses: step-security/harden-runner@6c439dc8bdf85cadbbce9ed30d1c7b959517bc49 # v2.12.2
+        with: { egress-policy: audit }
+      - uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
```
jasonkarns, line 19: FYI: pinning to shas is the recommended approach for security (because branches and tags are mutable and could be compromised via force pushes).

the trailing comment with version number _IS_ bumped by dependabot so it will stay accurate.

jasonkarns (commented):
self review for comments!

.github/workflows/test.yml
```diff
@@ -2,33 +2,23 @@ name: Tests
 
 on:
   push:
-    branches: [main]
   pull_request:
-    branches: [main]
+  workflow_dispatch:
+  schedule: [ cron: "42 0 7 * *" ] # https://crontab.guru/#42_0_7_*_*
```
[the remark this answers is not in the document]
jasonkarns: They do trigger a notification, yes. Though it's treated the same as other GHA notifications and subject to your notification settings.

image

.github/workflows/test.yml
```diff
@@ -2,33 +2,23 @@ name: Tests
...
-        ruby-version:
-          - "3.1"
-          - "3.2"
-          - "3.3"
-          - "3.4"
-
-    runs-on: ${{ matrix.os }}
-
+    runs-on: ubuntu-latest
+    strategy: {matrix: {ruby: [ '3.0', 3.1, 3.2, 3.3, 3.4 ]}}
```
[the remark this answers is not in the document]
jasonkarns: Yep. Our position might change if we do anything that isn't supported on 3.0. Or if rubocop drops support. But since we're a static tool that folks likely would be using as _part of_ upgrades (for autofixing and normalizing on new features, etc), I think it makes sense to have as broad a test base as possible. Until there's a reason to drop, anyway.

.github/workflows/test.yml
jasonkarns: (there is some irony, however, that even the shas are not hack-proof. to my knowledge, if someone were to push a ref that collides with a sha, github would use the ref, not the sha. 😱 )

PR #743, opened 2025-08-05
Remove devdep version specs from Gemfile

jasonkarns:
Version spec pinning dates back to their addition to the Gemfile. The version specs shouldn't really be necessary.

(Simplecov was separate because it used to be guarded with a ruby version conditional. That's no longer necessary so it shouldn't be isolated on its own.)


czeise (approved):
:shipit:

PR #778, opened 2026-01-07
Change required rubocop version: ~> 1.81.7 to ~> 1.82.1

r7kamura:
In order to use standard gem in ruby 4.0, I want to make it possible to use rubocop 1.82.0 or higher, which supports ruby 4.0. Therefore, following past conventions, I have proposed upgrading the rubocop version here. 

- Resolve https://github.com/standardrb/standard/issues/777

I referred to the following PR, as it seems to be where the previous change was implemented. 

- https://github.com/standardrb/standard/pull/767

While `~> 1.82.0` would have been sufficient, the latest rubocop version currently released is 1.82.1. Since previous version updates seem to have been based on setting the requirement to the latest version available at the time, I have set it to `~> 1.82.1` for this instance.

FYI: It looks like there's another related PR:

- https://github.com/standardrb/standard/pull/776

It appeared that this Dependabot PR was causing CI to fail because new Style/ModuleMemberExistenceCheck cop was added in version 1.82.0, so I have included the necessary fix in this PR. Following past conventions, I simply disabled it this time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## Unreleased
+
+* Updates rubocop to [1.82.1](https://github.com/rubocop/rubocop/tree/v1.82.1)
+
 ## 1.52.0
 
 * Updates rubocop to [1.81.7](https://github.com/rubocop/rubocop/tree/v1.81.7)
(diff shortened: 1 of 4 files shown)


r7kamura:
Yes, I agree. Thanks for adding that 👍 
In this pull request, I'm not aiming to complete everything up to Ruby 4.0 compatibility; instead, I'm focusing specifically on updating RuboCop as a single small unit of work.

